PR #14, opened 2022-07-07
Header and Footer styling

Mirzero:
An 80% implementation of the header and footer structure and styles.
I want to merge this down now so it's available for demo prep.


infojunkie (approved):
It's difficult to test this because it uses content blocks that are not being recreated locally. I was able to import the full config changes without problems though. Please go ahead and merge it.

PR #16, opened 2022-07-08
Transitioned the footer (and the Alert Ribbon) to use Simple Blocks so that the content is stored in the repo as configuration

Mirzero:


rjkuyvenhoven:
Looks good.  However, the image for the BC Logo in the footer is missing. `/sites/default/files/inline-images/footer-bc.png`  
We should be to link to the theme image.  'src="/themes/custom/workbc/assets/images/footer-bc.png"'

PR #24, opened 2022-07-12
[WR-268] Add error handling to SSoT query

infojunkie:
This PR adds some error handling to SSoT API query:
- A short "ping" at the beginning of the node view to check whether the API is up 
- A log message and NULL return in case of error

It is assumed that if the API is up, the subsequent call will not fail. 

To test: Change the `settings.local.php` value `ssot_url` to some other URL and attempt to display a career profile.


rjkuyvenhoven:
Tests gave the expected results.  Looks good.

rjkuyvenhoven (approved):

PR #29, opened 2022-07-19
[WR-390] SSoT source fields

infojunkie:
This PR adds new SSoT extra fields for the source information as per the wireframes. The source information is retrieved from a new SSoT endpoint `sources` that is keyed by the name of the endpoint used to retrieve the data. 

To test:
- Restore the latest SSoT dump and ensure a `ssot.sources` table has been created
- Sync the Drupal config
- Visit any career profile page and notice the various "Source:" fields


rjkuyvenhoven (approved):
I have tested and reviewed, source fields are displaying has expected.

PR #30, opened 2022-07-19
Wr 376 simple map with labels

rjkuyvenhoven:
Static map with region labels has been added to the Career Profile page.

Map should display on the Labour Market Statistics tab twice under Employment by Region and Job Openings


infojunkie (approved):
- Would have been good to get a higher res image from the design team
- At some point we will need to do an accessibility review
- I know it's overkill here, but I would like to float the idea of placing ALL HTML output in twig templates, to make it easier for future Drupal themers to adjust the HTML to suit their needs

In the interest of keeping the pace, let's go with what's here for now.

PR #31, opened 2022-07-20
[WR-394] create automatic url paths

rjkuyvenhoven:
I have created pathauto setting for each content type.
http://workbc.docker.localhost:8000/admin/config/search/path/patterns

most are straightforward. 

I didn't include the date as part of the path for blog and news as they don't seem to generate a significant number of blog/news posts.  

Event -  I included the date as part of the path.

Labour Market Monthly - I included the year and month as part of the path.

BC Profile and Region Profile share the same pattern.

### DIFF
--- a/src/config/sync/pathauto.pattern.basic_page.yml
+++ b/src/config/sync/pathauto.pattern.basic_page.yml
@@ -5,17 +5,18 @@ dependencies:
   module:
     - node
 id: basic_page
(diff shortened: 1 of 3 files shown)


infojunkie (approved):
Thanks, I tested by bulk-generating all paths from the Amdin UI at http://workbc.docker.localhost:8000/admin/config/search/path/update_bulk

PR #32, opened 2022-07-25
Video Library listing theming

Mirzero:
This doesn't include theming for the 'YouTube window' which is not yet implemented.


infojunkie:
> YouTube window

This is a Lightbox-style interaction, which is all front-end based. Since you are manipulating the view to format the video card HTML, can you suggest an addition there that implements the Lightbox interaction?

Mirzero:
> > YouTube window
> 
> This is a Lightbox-style interaction, which is all front-end based. Since you are manipulating the view to format the video card HTML, can you suggest an addition there that implements the Lightbox interaction?

The view just spits out an a tag... but it would need to emit a bootstrap modal or something like that instead, with the link connected to toggling that modal. The rendering of the a tag is not part of the view template, it's part of the field template (field: Media Thumbnail) and I'm not sure of the optimal way to get Drupal to embed the necessary modal as part of that field rendering

infojunkie:
Yes, I think @rjkuyvenhoven should change the view so that both thumbnail and title both link to the YouTube video - which you can then target for Lightbox.

PR #39, opened 2022-07-27
Blog - Implemented Template and Styles

Mirzero:


rjkuyvenhoven:
When only two blog items are being shown, the first is placed in the left column, the second is placed in the right column and the middle column is empty.  Other than that, looks good.

PR #40, opened 2022-07-27
Success Stories - Implemented Template and Styles

Mirzero:


rjkuyvenhoven:
When there are only 2 items on a row the are located in the left and right column with the middle column remaining empty.  Otherwise, looks good.

PR #41, opened 2022-07-27
[WR-357, WR-395] KeyCloak and config_ignore

infojunkie:
This PR installs the KeyCloak module and fixes the Drupal config infrastructure to ignore the KeyCloak settings. This includes:

- Install KeyCloak module and dependencies
- Remove config_split module (as per WR-395) and remove `src/sync/split` files
- Enable config_ignore and ignore KeyCloak and Backup / Migrate settings
- Replace config_split settings in settings.php with the simpler approach of excluding specific modules in non-dev environments
- Remove redundant environment variables


rjkuyvenhoven:
Attempting to sync results in the following errors

```
 [error]   (Currently using Missing or invalid module The following module is marked as installed in the core.extension
configuration, but it is missing:
 * config_split
```

`  The "config_split" entity type does not exist.`

manually installing the config split module followed by executing the `drush cr` and `drush cim` commands get everything up and running.


we may need to do this in two stages.  First do all the configuration changes without uninstalling the `config split` module.  Once everyone (or at least AWS-DEV) is on that master with the new config, then uninstall the module.

infojunkie:
Thanks for the review. I'll try again with a dump from live to test the workflow.

infojunkie:
I was able to reproduce with a dev dump. I solved it using the following:
```
drush cdel core.extension module.config_split
drush cr 
```
then the sync script passed. Thinking about how to make this productionizable.

infojunkie:
Pushed a small hack to force-delete the config_split module and keep going with sync. @rjkuyvenhoven, please try again. Thanks!

rjkuyvenhoven (approved):
I was able to successfully sync to this branch.

infojunkie:
Thanks for the confirmation. However, I realized that this will not work for AWS because we run Composer during the Docker build phase, not at runtime. There's no Drupal environment that drush can find during the build phase. I am pausing this branching and decoupling Keycloak from config_split.

infojunkie:
I was able to perform the following workflow successfully on my local:
- Restore the latest dump from DEV 
- Uninstall the config_split module from the UI
- Run sync successfully
If @rjkuyvenhoven doesn't mind, I will call this a success and merge it to `master` as is. Local devs will not need to manually uninstall the module because the Composer post-install script takes care of that already as you previously confirmed.

rjkuyvenhoven:
ok, sounds good to me.

PR #42, opened 2022-07-29
WR-435 - Related Topics Component - Display template and Carousel implementation

Mirzero:


rjkuyvenhoven:
The styling looks good.

When I click the next icon the images slide off to the left leaving blank space and then the next set of images slide in from the right.

PR #43, opened 2022-08-02
[WR-389] Upgrade Drupal and config

infojunkie:
This PR upgrades dependencies:
- Drupal goes to 9.4.4 
- Other dependencies are upgraded
- `config_filter` is kept at version `~1` (don't move to 2.x) because of inconsistency in config ignore results
- Config sync is upgraded accordingly
- `time_field` patch is added


rjkuyvenhoven (approved):
Once I increased the process timeout value I was able to successfully upgrade.

PR #46, opened 2022-08-04
WR-435 - Related Topic Block - Moved the header into the block so that it can be hidden when there are no related topics

Mirzero:


rjkuyvenhoven (approved):
Works as expected.

PR #47, opened 2022-08-08
WR-370 - Publication theme

Mirzero:


src/web/sites/default/settings.local.php
```diff
@@ -168,3 +168,9 @@
  * WorkBC configuration.
  */
 $config['workbc']['ssot_url'] = 'http://ssot:3000';
+
+ini_set('memory_limit', '6G');
```
infojunkie, line 172: What is the resource that needs this much memory?

PR #58, opened 2022-08-15
WR-508: Fixed styling issues with the hero banner

Mirzero:


zetagraph (approved):
@Mirzero LGTM 

Wondering if we need to keep the commented sections in both TWIG and SCSS files? Or maybe we can clean that up right away?

I would also lean toward not nesting SCSS too much. I usually only nest the media queries and pseudo-elements. I think it's easier to read and creates less specificity in compiled CSS.

We can hopefully have more control over markup and use BEM naming conventions down the line.

Mirzero:
Ah good catch - I didn't mean to leave any of that commenting in there.

PR #62, opened 2022-08-18
WR-507: Global Webform styling, and some bad style cleanup

Mirzero:


src/web/themes/custom/workbc/scss/abstracts/_variables_drupal.scss
```diff
@@ -11,13 +11,13 @@ $navbar-brand-site-title-font-weight: bold !default;
 $navbar-brand-site-title-letter-spacing: 2px !default;
 $navbar-brand-site-slogan-font-size: .875rem !default;
 
-$region-padding: 0 .5rem !default;
+$region-padding: 0 8px !default;
 
-$footer-padding: 2rem 0 !default;
+$footer-padding: 30px 0 !default;
```
zetagraph, line 16: Don't think this is used anywhere now. Footer has the styles in its partial. 

I like consistency but choose flexibility over it. That's why I prefer not to abstract things into variables too much.

zetagraph (approved):
@Mirzero LGTM

PR #63, opened 2022-08-18
WR-403: Responsive Menu

zetagraph:
Tickets:

- Responsive / Off Canvas Nav
- "Mmenu" for nested level navigation
- Dependency Libraries 
- Initial Styling

This will need additional tweaks and updates that I will be adding in subsequent passes. Structurally things should be there. though.


Mirzero (approved):
It seems to work well! I see a few very minor style things, but I think this is more about the module and functionality, and that all seems great to me.

PR #64, opened 2022-08-19
WR-526: Implemented a left-nav block with placeholder content (real HTML structure) and styling

Mirzero:


infojunkie (approved):
Looks good to me. We will need to find a way to target specific pages somehow at a later stage.

zetagraph (approved):
@Mirzero LGTM

PR #66, opened 2022-08-23
WR-403: Mobile and global theme updates

zetagraph:
Tickets: 


src/web/themes/custom/workbc/templates/navigation/menu.html.twig
```diff
@@ -0,0 +1,55 @@
+{#
```
infojunkie, line 1: Where is this menu template used? Note that we have a block that generates the desktop navigation menu - should / can these two pieces of code be merged or at least co-located?

zetagraph: This is used on all the other menus except the homepage so that they don't inherit the default Bootstrap stuff. This is an instance where we have to fight the Bootstrap although we would've probably ended up having this template anyway. Don't think this can be combined with the Homepage menu as it's a custom setup.

infojunkie: ok sounds good, thanks.

infojunkie (approved):
Lgtm

PR #67, opened 2022-08-23
Remove landing page content type

rjkuyvenhoven:
1. download backups of both "Drupal Default Database" and "Public Files Directory" from AWS-DEV
2. restore these two backups to your local
3. run the sync script

restoring backup from AWS-DEV is required to ensure there is no "Landing Page" content prior to the updated configuration being imported.


infojunkie (approved):
Worked for me locally, thanks

PR #73, opened 2022-08-31
WR-364: Theme full width image cards

zetagraph:
Tickets: https://hive.aved.gov.bc.ca/jira/browse/WR-364


rjkuyvenhoven:
Since we only need a single indicator for left or right side start for a group of Full Width action cards.  I think it makes more sense to put the indicator in the **Action Cards: Full Width** (action_cards_full_width) paragraph. Rather than creating a new full width action card.

zetagraph:
@rjkuyvenhoven the thing is that these cards don't always show in groups. For example here:

http://aved.number41media.com/Wireframes/design-refresh/Beta2022/#id=a8iqb9&p=l1_-_search___prepare_for_a_job_-_approved&g=1

or here: 

http://aved.number41media.com/Wireframes/design-refresh/Beta2022/#id=t0aahk&p=l3_-_resources_for_parents_-_approved&g=1

or here: 

http://aved.number41media.com/Wireframes/design-refresh/Beta2022/#id=mj73q4&p=l3_-_find_your_fit_tour_-_old&g=1


They are just a single card with the left or right alignment. 

I know we are still using a group paragraph but thought it would be more flexible to have the option on the individual card. 

We are creating individual templates for these paragraphs anyway. I feel that having things in separate paragraphs rather than display modes adds more flexibility. The config updates are not that big of a change in my opinion, the same fields are reused.  

I have things themed and working locally:

Screen Shot 2022-08-31 at 1 44 41 PM

Screen Shot 2022-08-31 at 1 43 56 PM

rjkuyvenhoven:
If we take this approach then what happens for the second card?  Is the alignment field ignored? or is the editor required to select the appropriate alignment?

zetagraph:
@rjkuyvenhoven cards start aligned to the left by default. If an editor wants to alternate, the second card is aliged to the right. This also opens up the possibility of having all cards aligned to the left or to the right. That possibly might be the desired option at some point. 

For example this here:

http://aved.number41media.com/Wireframes/design-refresh/Beta2022/#id=8xc7m2&p=l3_-_resources_for_women_-_approved&g=1

Although not sure yet if we would reuse this component or its variation for this style of card.

rjkuyvenhoven:
I think this overcomplicates things.   This will also have an impact on content migration so we'll need to run this by Karim.

http://aved.number41media.com/Wireframes/design-refresh/Beta2022/#id=wuwi4b&p=l1_-_research_the_labour_market_-_approved&g=1

If an editor wanted to add a Feature Card to this page, the editor would also have to change the alignment setting for each card.  Or if the editor wanted to re-arrange the order of the cards.

 
 By adding the alignment field to the Action Cards: Full Width paragraph instead, we don't have to worry about this. In addition, if in the future we wanted to handle all cards aligned the same we could add "All Left" and "All Right" options to the alignment field and adjust the theme accordingly.

infojunkie:
I also prefer to have the alignment specified in the container, not at the card level. If we want other alignment options, we can add more values to the list, e.g. "all left" or "all right". I am assuming there will always be a container.

zetagraph:
Ok, the alignment option is moved to the container now. 

We now need a way to pass the "Alignment key" class from the Container paragraph (Full-Width Cards) to the cards themselves (Full-Width Card) 

It needs this class (cards-w-image--left or card-w-image--right) for proper aligment/theming 

Screen Shot 2022-08-31 at 5 26 37 PM

@rjkuyvenhoven can you please assist with passing that variable?

rjkuyvenhoven:
Sure. 

You need the cards  to alternate between card-w-image--right and card-w-image--left, starting with the appropriate class depending on the alignment variable?

PR #77, opened 2022-09-07
WR-364: Theme Cards and Layouts

zetagraph:
Tickets: 


infojunkie (approved):
Looks good to me. But this does nto cover all card types, correct?

Mirzero (approved):
looks good to me

zetagraph:
@infojunkie these should cover all the "card in columns/rows" layouts. Two, three, four per row. We still to address the Quote card and the special case cards, e.g. Publications, Explore Careers. I think that's it.

PR #81, opened 2022-09-09
WR-435: Related Topics Carousel - reimplement with Swiper js

Mirzero:


zetagraph (approved):
@Mirzero this looks good to me. 

There is an option to put the JS files into the "libraries" directory and serve that locally instead of using a CDN but don't know if that would necessarily be of benefit here.

You can see an example of the Library install/include for the mobile menu JS here:

https://github.com/bcgov/workbc-main/blob/cdd73d67e64082cea8c1abb1c492eaf2857657c0/src/composer.json#L40

PR #84, opened 2022-09-12
WR-277: Added content scraping and link and anchor generation

Mirzero:


src/web/themes/custom/workbc/js/sidenav-anchors.js
```diff
@@ -0,0 +1,35 @@
...
+    var anchorid = "sideNavAnchorId_" + index;
+    var anchorText = jQuery(element).text();
+
+    $(this).wrapInner('<a id="' + anchorid + '" />');
+
+    linkList.append('<li><a href="#' + anchorid + '">' + anchorText + '</a></li>');
+  };
+
+  let initSidenavAnchors = function() {
+    console.log('Begin sidenav configuration');
```
infojunkie, line 18: should probably remove those debugging logs before merging.

infojunkie (approved):
Love the simplicity of the implementation. My only concern would be to ensure that h2 is actually the title size that the CKEditor produces by default.

PR #86, opened 2022-09-13
WR-663: Related topic layout updates

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-663


zetagraph:
@rjkuyvenhoven 

This is the warning I was referring to. Comes up on some pages but not all of them. 

Screen Shot 2022-09-13 at 2 42 09 PM

zetagraph:
@rjkuyvenhoven is there a way to not render an empty block if there are no related items? It is already wrapped in the `if` statement. But assuming the block is triggered because it has something still? We just need to get rid of the separate line that is still part of the block styling and show up on pages even if the block is empty (no related items).

 
Screen Shot 2022-09-13 at 4 47 34 PM

rjkuyvenhoven:
I've committed an update that should resolve this problem.

zetagraph:
Thanks, @rjkuyvenhoven Merging.

PR #88, opened 2022-09-15
[WR-378] Adds Events list and calendar views

anuragparihar1:
This piece adds two views and a custom module (i.e. calendar_listview) 
Two new paths should show Calendar (i.e. /calendar/view/calendar) and list (i.e. /calendar/view/list) view.


infojunkie (commented):
I don't think we want the folder `css/backup-15-09-2022`

src/config/sync/field.field.node.event.field_target_audience.yml
```diff
@@ -0,0 +1,19 @@
...
+entity_type: node
+bundle: event
+label: 'Target Audience'
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: string
```
infojunkie, line 19: We already have a field `field_audience` and a taxonomy `target_audience` - can @rjkuyvenhoven advise which one we should be using here?

infojunkie (commented):
`src/web/modules/custom/calendar_listview/css/cv_custom.css` is empty - is it needed?

infojunkie (commented):
We don't want to include boostrap and jquery as dependencies in this custom module - would it be possible to reuse the global ones?

src/web/modules/custom/calendar_listview/js/cv-custom.js
```diff
@@ -0,0 +1,91 @@
...
+          if($('.js-drupal-fullcalendar div.pager2').length <= 0){
+            $(".fc-view-container").once().after("<div class='pager2'></div>");
+            $(".js-drupal-fullcalendar div.pager2").once().html(pagerCalendarView);
+            $(".js-drupal-fullcalendar div.pager2 .fc-center").hide();
+          }
+        }
+        
+        var dcv = $(".js-drupal-fullcalendar .fc-content-skeleton");
+        
+        if(typeof dcv.attr('class') !== "undefined"){
```
infojunkie, line 38: If this block does nothing, we can remove it.

src/web/modules/custom/calendar_listview/js/_cv-custom.js
```diff
@@ -0,0 +1,31 @@
+(function ($, Drupal) {
+  Drupal.behaviors.cv = {
```
infojunkie, line 2: Is this file current? There's another file named `cv-custom.js` (without leading undersscore)

src/config/sync/field.field.node.event.field_target_audience.yml
rjkuyvenhoven: `field_audience ` is the field we should be using.

The `field_target_audience` field should no longer exist in the events content type.  This field was removed in a commit I made to master late last week.

rjkuyvenhoven:
I added several events to take place in September. These events don't show up when I look at the Calendar View.  Events that were entered previously do show.  The new events do show on List view.

src/web/modules/custom/calendar_listview/calendar_listview.info.yml
```diff
@@ -0,0 +1,11 @@
+name: Calendar Listview
```
infojunkie, line 1: This module is not enabled in this branch. It should be enabled in the config sync.

rjkuyvenhoven: In addition, Anonymous and Authenticated user should given the permission to view the Calendar and List View

infojunkie:
I checked out this branch, ran the sync, enabled the custom module, but http://workbc.docker.localhost:8000/calendar-upcoming-events shows me a blank page even after adding an event. 

@anuragparihar1 can you please add installation and test instructions here, to help review the PR?

infojunkie:
> I added several events to take place in September. These events don't show up when I look at the Calendar View. Events that were entered previously do show. The new events do show on List view.

I added an event that showed in the calendar view. Maybe there is a condition that the view misses? Please show us the events that were missed.

src/web/modules/custom/calendar_listview/calendar_listview.info.yml
infojunkie: @anuragparihar1 I have added the module to the config sync.

infojunkie:
Please package your views as blocks to be injected inside the page `/calendar-upcoming-events`. The reason for this is to allow the content editors to insert a hero banner and image. @rjkuyvenhoven can explain the mechanics of injecting a view block to a target page.

rjkuyvenhoven:
After deleting previously existing events (which were generated by Devel Generate) the new events showed just fine.  

The content generated by Devel Generate is suspect as data is random and the result is date and time  ranges that don't make sense and it appears this was confusing the view.

rjkuyvenhoven:
Calendar-View-WorkBC

The dash should not appear for single day events.

rjkuyvenhoven:
Some custom validation will need to be added to the Event edit form.

End date & time must be greater than the Start date & time
End date is only necessary for multi-day events so it should be empty or greater than the Start date.

src/config/sync/field.field.node.event.field_target_audience.yml
infojunkie: @rjkuyvenhoven what about the taxonomy `target_audience` ? Is it being used at  all?

zetagraph (commented):
@anuragparihar1  

Wondering if should be doing all the styling in the module instead of the theme. Seems like we are duplicating a lot of Bootstrap stuff, SCSS, mixins, and assets that already exist in the theme. 
Screen Shot 2022-09-16 at 1 40 30 PM

For example, this style already exists in the theme as a `mixin` and there is no need to load the image again.

Screen Shot 2022-09-16 at 1 35 23 PM

Noticed a duplication of links locally. Not sure if it's just the lack of content. 
Screen Shot 2022-09-16 at 1 42 49 PM

Things look good in general. There are lots of moving pieces. Thanks for the work.

src/config/sync/field.field.node.event.field_target_audience.yml
rjkuyvenhoven: The `target_audience` Taxonomy can also be removed. It is no longer needed.

infojunkie:
Thanks @anuragparihar1 for these changes, code looks much cleaner now. 

There are a few unanswered comments above:
- The event validation reported by @rjkuyvenhoven 
- The edge case reported by @zetagraph of empty list view showing top and bottom next/previous links stacked together
- The duplication reported by @zetagraph about the image mixin

anuragparihar1:
Thanks for your input guys. Truly appreciated

@rjkuyvenhoven Re Event validation: This task was seen as List & Calendar views creation. Event node validation can be added in the custom module if this appears to be the best place for it. Please confirm.

 @zetagraph Re empty list view: Yes, this is due to the lack of content, and 
Re Image mixins: I was, and still am unaware of the full list of SVGs created. I can go in and replace them, but finding the list of the existing ones and matching them with the ones used would be a time taking part. It can be done if we see that we want to invest time into it. For future tasks, I'll certainly look into the available mixins first.

rjkuyvenhoven:
> @rjkuyvenhoven Re Event validation: This task was seen as List & Calendar views creation. Event node validation can be added in the custom module if this appears to be the best place for it. Please confirm.

Yes, added in the custom module is the best place.  NOTE: I've just merged my workflow branch into master which contains updates to the custom module.

zetagraph:
@anuragparihar1 Not concerned so much about the reuse of the mixins but overall styles separation and duplication in a module. Mentioned here: https://github.com/bcgov/workbc-main/pull/88#pullrequestreview-1111036439

anuragparihar1:
@rjkuyvenhoven Validation for date and time has been added. It tests that the start date is NOT smaller than the end date and if both dates are the same, then checks that the end time is NOT smaller than the start time.

infojunkie:
Thanks @anuragparihar1 if nothing else is pending here you can go ahead and merge squash.

zetagraph:
@anuragparihar1 @infojunkie what about the styles duplication are we doing anything about those?

https://github.com/bcgov/workbc-main/pull/88#pullrequestreview-1111036439

anuragparihar1:
@zetagraph We are using Bootstrap from the theme now. Some minor design elements could be duplicated, but most of the styling is for the calendar view. I'd suggest pushing this and getting it tested by our QA team.

PR #89, opened 2022-09-15
WR-634: search theming updates

zetagraph:
Tickets:

https://hive.aved.gov.bc.ca/jira/browse/WR-634


infojunkie (approved):
Lgtm

src/web/themes/custom/workbc/scss/components/_search-form.scss
```diff
@@ -0,0 +1,83 @@
...
+  &:focus {
+    outline: none;
+    box-shadow: none;
+  }
+}
+
+// Header Search
+.header__search .search-form .form-control,
+.header__search .search-form .form-submit.btn-primary {
+  background-color: #f2f2f2;
```
Mirzero, line 58: I prefer to use the color variables set up in _variables.scss, but perhaps it's a pipe dream that we'll ever be able to change one color in there and have it cascade across the site in exactly the way we want. I don't think I've ever had a project actually change one color like that, it's always a full redesign if it happens.

Mirzero (approved):
Looks like this cleans up a lot of things with how it was structured before, which is great

src/web/themes/custom/workbc/scss/components/_search-form.scss
zetagraph: Updated. 

Agree about the "use across" concept but at the same time a variable that is 32 characters long and simply represents`#fff`; doesn't seem to be much help to me. 

Also, every time I want to simply use a white colour I have to go and search for this abstraction hidden in another place. And it's hard to actually tell by reading something like this `$workbc_palette_light_primary` that it would actually represent white. 

And if we are talking about optimizing even further should we have everything prefixed with `workbc_` we are already in the context of "workbc" no?

Just my take on things. The naming of stuff and various code conventions can be talked about forever. Fine by me either way.

PR #92, opened 2022-09-20
WR-344: Main Nav updates and additions

zetagraph:
Tickets:


src/web/themes/custom/workbc/scss/layout/_nav_main.scss
```diff
@@ -71,7 +69,8 @@
             width: 50%;
 
             & > li {
-              border-left: $workbc_trim_standard_size solid $workbc_palette_transparent; //placeholder to prevent content shifting when colored borders are in place (hover, etc.)
+              border-left: $workbc_trim_standard_size solid
+                $workbc_palette_transparent; //placeholder to prevent content shifting when colored borders are in place (hover, etc.)
 
               & > a {
```
zetagraph, line 75: @Mirzero would really love if we can unnest this. We do need to be that specific with the CSS (depth) since we have control over the markup. Plus it's very hard to read. I would only use Nesting for Medi Queries and Pseudo elements.

Mirzero (approved):
Looks good to me. I didn't spin it up locally to look at the render, but I think this is all stuff we need regardless, and can tweak if there are any render issues.

infojunkie:
Minor point: is this WR-344 or WR-364?

src/web/themes/custom/workbc/scss/layout/_nav_main.scss
Mirzero: Are you asking me to do that, or asking if I'll be annoyed if you do it?

zetagraph: I think it's fine for now. Definitely not a priority but if you are in that context at some point might be good to clean up some classes and nesting. And I meant "We do not need..." in the comment above.

PR #94, opened 2022-09-22
WR-364: Theme layouts restructure

zetagraph:
Tickets: 


zetagraph:
@Mirzero @infojunkie

There is quite a bit of restructuring here. Trying to remove the reliance on Drupal Blocks and Views in content and template rendering. It will give us a lot more flexibility in terms of markup and structure in various content types/nodes. 

Dave, I've set up a Career Profiles node template and can guide you through the setup if needed. But basically the `node--page--full.html.twig` can be viewed as an example. 

It might be a good idea to check this branch out and test things locally. I am removing a bunch of blocks and unneeded templates. Will also remove Hero and Title Views once we confirm that all is in order.

There are a few more updates that will be needed for news and success stories. Work in progress.

infojunkie:
At first glance, this looks good. The lines to replace the hero views with simple if conditions are simple, as expected. I will leave it to @Mirzero to test the actual changes locally.

src/web/themes/custom/workbc/templates/layout/page--4xx.html.twig
```diff
@@ -0,0 +1,65 @@
+{#
```
Mirzero, line 1: I'm unclear what this twig is for, and what it's going to be messing with. --4xx? What page would that be, anything in the nodes with an id of 400-499?

Mirzero (approved):
Looks like a much cleaner structure than we had.

src/web/themes/custom/workbc/templates/layout/page--4xx.html.twig
zetagraph: This is for the 404 pages. We would need to address a few things that don't fall under the node layout options.

PR #101, opened 2022-09-27
WR-364: Layout blocks views updates

zetagraph:
@Mirzero @infojunkie had to shuffle a few things again in the attempt to decouple some views from the "page display" and move them (views) to "blocks" which can be dropped in pages. Only partially successful with that. The Search pages can't be used as blocks so left those as is with some template overrides. 

@infojunkie There is still some search duplication happening from what I can tell. Perhaps we can discuss this in the next Dev sync. 

@Mirzero There are still a few things that need to be realigned with the view templates you set up earlier. I think the "definitions" view page can be moved to a block but I wasn't sure about some functionality. Will confirm with you at some latter point.


infojunkie (approved):
Assuming that no functionality is lost or broken due to this refactoring, I approve.

PR #102, opened 2022-09-28
[WR-753] Creates a configurable block of Recent Jobs panel to be displayed on …

anuragparihar1:
…Career Profile content type.

Steps to make it work
01. Enable the module workbc_jobboard
02. Configure the Recent Jobs block (Along with the other settings please select Content Type to be Career Profile under visibility)
3.  Test on any Career Profile page


infojunkie:
Thanks @anuragparihar1. We can test but then please ensure to add the config export to enable the module and block placement.

src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
```diff
@@ -0,0 +1,119 @@
...
+      $form_state->setErrorByName('job_board_sub_title', $this->t('Sub Title can\'t be empty.'));
+    }
+    if (empty($form_state->getValue('job_board_results_to_show'))) {
+      $form_state->setErrorByName('job_board_results_to_show', $this->t('No. of results field can\'t be empty.'));
+    }
+  }
+	/**
+	*{@inheritdoc}
+	*/	
+	public function build(){
```
infojunkie, line 86: Please fix indentation here and across the PR.

src/web/modules/custom/workbc_jobboard/config/config.ini
```diff
@@ -0,0 +1,20 @@
+[ENVIRONMENTS]
+WWW           = 'workbc.ca'
+UAT           = 'aws-uat.workbc.ca'
+DEV           = 'aws-dev.workbc.ca'
+LOCALHOST     = 'workbc.docker.localhost:8000'
+
+[WS-HOSTS]
+WWW           = 'https://user:pass@api-jobboard.workbc.ca'
+UAT           = 'https://user:pass@api-jobboard.workbc.ca'
```
infojunkie, line 9: We don't want to store the username/password in the repo. 

To maintain a consistent approach of configuring services, I suggest to remove the ini file completely and instead use a config setting such as `ssot_url` https://github.com/bcgov/workbc-main/blob/master/src/web/sites/default/settings.local.php#L170 to determine the Job Board URL. The same setting is obtained in non-local stages via an environment variable https://github.com/bcgov/workbc-main/blob/master/src/web/sites/default/settings.aws.php#L23 

This should also remove the `fnGetService()` function since the UrL will be given explicitly by the config setting above.

src/web/modules/custom/workbc_jobboard/src/Controller/apiServices.php
```diff
@@ -0,0 +1,153 @@
...
+		  if(!empty($cookie_vals)){
+			$cookie = $cookie_vals;
+		  }
+		  else
+		  {
+			$cookie = NULL;
+		  }
+		}
+    
+    if($url){
```
infojunkie, line 59: Please consider using Guzzle which is shipped with Drupal by default and is higher-level than curl, e.g. https://github.com/bcgov/workbc-main/blob/master/src/web/modules/custom/workbc_extra_fields/workbc_extra_fields.module#L71-L87

src/web/modules/custom/workbc_jobboard/src/Controller/apiServices.php
```diff
@@ -0,0 +1,153 @@
...
+            $r['cookies'] = $m;
+          }
+          $r['response'] = $ret;
+          
+          return $r;
+        }
+        return false;
+		  }
+		  catch (Exception $e) {
+        return false;
```
infojunkie, line 139: Please log exceptions to `\Drupal::logger` instead of silently ignoring them.

infojunkie (changes requested):
Thanks @anuragparihar1 for this PR! Looking great so far. I've left a few comments that should hopefully allow us to remove some code here and remain consistent with the rest of the app.

src/web/modules/custom/workbc_jobboard/config/config.ini
```diff
@@ -0,0 +1,20 @@
...
+LOCALHOST     = 'https://user:pass@api-jobboard.workbc.ca'
+
+; To test with local JSON file: 
+; 01. Change WS-HOST for LOCALHOST to the environment name (where local JSON file is saved) and 
+; 02. Uncomment getRecentJobsJson() method in WorkBcJobboardController, and 
+; 03. Uncomment workbc_jobboard.get_recent_jobs routing under workbc_jobboard.routing.yml.
+
+
+[SERVICES]
+GET_RECENT_POST       = 'api/career-profiles/topjobs'
```
infojunkie, line 20: I don't think this needs to be a setting. A `const` in the API file should be sufficient.

src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
```diff
@@ -0,0 +1,119 @@
...
+ *   category = @Translation("Workbc Jobboard Sidebar"),
+ * )
+ */
+ 
+class WorkbcJobboardSidebar extends BlockBase{
+  
+  /**
+   * {@inheritdoc}
+   */
+  public function blockForm($form, FormStateInterface $form_state) {
```
infojunkie, line 27: Thanks for making the block configurable!

anuragparihar1:
@infojunkie Your suggestions have been implemented. Please review when you can

src/web/modules/custom/workbc_jobboard/workbc_jobboard.routing.yml
```diff
@@ -0,0 +1,9 @@
+# workbc_jobboard.get_recent_jobs:
```
infojunkie, line 1: Do we need this file if it's empty?

src/web/sites/default/settings.aws.php
```diff
@@ -21,3 +21,5 @@
 
 // Single Source of Truth (SSoT) configuration.
 $config['workbc']['ssot_url'] = getenv('SSOT_URL');
+
+$config['jobboard']['api_url'] = getenv('api_url');
```
infojunkie, line 25: Sorry to be annoying, but let's please keep the environment variables upper-case.

infojunkie (changes requested):
The code is looking much simpler, thanks!

1 minor change needed, one file to delete, and we're good to go.

src/web/modules/custom/workbc_jobboard/workbc_jobboard.routing.yml
anuragparihar1: @infojunkie Content of this file is commented out. This file is for testing purposes when you want to hit an internal JSON file instead of hitting the API.

src/web/sites/default/settings.aws.php
infojunkie: Also, please rename the environment variable something less generic, like `JOBBOARD_API_URL`. Thanks!

src/web/modules/custom/workbc_jobboard/workbc_jobboard.routing.yml
infojunkie: Thanks - a comment at the top might be useful to avoid this confusion.

anuragparihar1:
@infojunkie Suggested changes have been pushed.

infojunkie (approved):
Thank you! Good to go from my side.

PR #105, opened 2022-09-29
Adds fvm module and gives content editor ability to set viewmode of a…

anuragparihar1:
… page

Please check


infojunkie:
Why do we need to give editor ability to change the viewmode? If this is related to the calendar page theming, we only need to specify a different viewmode, but not expose this ability to editors.

anuragparihar1:
@infojunkie We need some way of telling Drupal which node would use the different viewmode.  [ fvm module](https://www.drupal.org/project/fvm) is used to change the viewmode. If needed we can either disable the field in the Form Display or not give the content editor access to this field. Suggestions?

infojunkie:
Right, I was just not expecting a full module needed for that. No need to expose that field to other than full admins.

anuragparihar1:
@infojunkie Field disabled in the Form Display to secure access.

infojunkie (approved):
Thanks, looks good to me now.

PR #107, opened 2022-09-30
WR-747: Theme updates fixes

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-747
https://hive.aved.gov.bc.ca/jira/browse/WR-748

### DIFF
--- a/src/web/themes/custom/workbc/templates/content/node--career-profile--full.html.twig
+++ b/src/web/themes/custom/workbc/templates/content/node--career-profile--full.html.twig
@@ -87,7 +87,8 @@
   {% include '@workbc/includes/hero.html.twig' %}
 {% endif %}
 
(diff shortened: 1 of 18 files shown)


infojunkie (approved):
Looks good to me, thanks

src/web/themes/custom/workbc/templates/content/node--career-profile--full.html.twig
```diff
@@ -87,7 +87,8 @@
   {% include '@workbc/includes/hero.html.twig' %}
 {% endif %}
 
-{% include '@workbc/includes/breadcrumbs.html.twig' %}
```
zetagraph, line 90: @Mirzero had to change how the block is rendered here. You will have a one-liner conflict here.

PR #110, opened 2022-10-04
WR-748: theme form updates

zetagraph:
Tickkets:

https://hive.aved.gov.bc.ca/jira/browse/WR-748


src/config/sync/block.block.webform.yml
```diff
@@ -36,4 +36,4 @@ visibility:
   request_path:
     id: request_path
     negate: false
-    pages: /career-profiles
+    pages: /plan-career/our-publications/workbc-order-form
```
infojunkie, line 39: 1. I don't understand how the path /career-profiles is related to the order form
2. Are we changing all paths in this PR? Otherwise, how did you make sure that this new path will in fact be the one that's applied to all other pages?

src/web/themes/custom/workbc/scss/components/_tabels.scss
```diff
@@ -0,0 +1,40 @@
+legend {
```
infojunkie, line 1: Type in filename (`_tabels` instead of `_tables`)

src/web/themes/custom/workbc/scss/style.scss
```diff
@@ -36,23 +35,26 @@
...
-@import 'components/card-career';
-@import 'components/cards-grids';
-@import 'components/alphabet-navigation';
-@import 'components/alphabet-search';
-@import 'components/request-tour-stop';
 @import 'components/social_sharing_buttons';
-@import 'components/post';
+@import 'components/video-library-modal';
+@import 'components/workbc-card';
+@import 'components/tabels';
```
infojunkie, line 60: Typo in filename

src/web/themes/custom/workbc/scss/components/_tabels.scss
zetagraph: corrected, thanks for spotting

src/config/sync/block.block.webform.yml
zetagraph: `/career-profiles` is incorrect path in this case. The form should be living under `/plan-career/our-publications/workbc-order-form` which would be a page that I will create on Dev

infojunkie: This page needs to be imported from the IA sheet, otherwise the content migration will be out of sync. Let's discuss this to get a better understanding of what's needed.

zetagraph: @infojunkie made webform config and block updates as per our conversation. Path aliases will be addressed in a separate PR

infojunkie (approved):
Lgtm thanks

Mirzero (approved):
Looks good to me

PR #116, opened 2022-10-05
WR-762: url aliases and block paths updates

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-762


infojunkie (commented):
Looks great, thanks! 

For the sake of completeness, can we also include the `/definitions` view as a block within a page? This is the last remaining view that does not work like the others.

zetagraph:
@infojunkie made several updates and changes. I had to keep the "Definitions" and  "Career Search A-Z" as pages. Upon further investigation with @rjkuyvenhoven , we didn't find a way to keep the "Views Attachment" functionality while converting things to a block. It has to be a page in order for the "letter search" do it's thing with the URL

infojunkie (approved):
Lgtm, thanks

PR #117, opened 2022-10-11
WR-361 Career Profiles theme

Mirzero:
This implements the majority of the theming for Career Profiles.
There are still some issues with theming the content within the tabbed sections, but that's not a new problem so I don't see a concern there with merging.

This merge introduces a new dependency on an external js library - ScrollBooster. This was needed to implement the scrollable tabs per wireframe design.


src/web/themes/custom/workbc/templates/content/node--career-profile--full.html.twig
```diff
@@ -70,7 +70,50 @@
...
+{% set statistic_annual_salary_source = content.extra_field_annual_salary_source %}
+{% set statistic_minimum_education = content.extra_field_minimum_education %}
+{% set statistic_job_openings = content.extra_field_job_openings %}
+{% set statistic_job_openings_source = content.extra_field_job_openings_source %}
+
+{% set career_profile_tabbed_content = content.group_career_profile_tabs %}
+
+{% set tab_groups = {
+  'item-1': {
+    'title': 'Career Overview',
```
infojunkie, line 87: Since we're using `some text | t` in some places, can we use it everywhere ?

src/web/themes/custom/workbc/templates/includes/social-sharing.html.twig
```diff
@@ -0,0 +1,7 @@
+<div class="social-sharing-wrapper">
+  <div class="container-lg">
+    <div class="social-sharing-content">
+        Social Sharing icons placeholder
```
infojunkie, line 4: Is this placeholder intentional here?

infojunkie (commented):
Code looks neat and well-organized, thanks! I haven't tested locally.

I added a couple of comments, but otherwise lgtm.

src/web/themes/custom/workbc/templates/content/node--career-profile--full.html.twig
Mirzero: Pushed an update to expand it's use

src/web/themes/custom/workbc/templates/includes/social-sharing.html.twig
Mirzero: For the moment, yes. It obviously needs to get filled, but those icons aren't in master yet either for this page as far as I know so it's not a regression. It's on the list of things to fix later though.

infojunkie (approved):
Lgtm, thanks!

PR #118, opened 2022-10-12
WR-364: updates fixes tweaks

zetagraph:
Various Tickets and Bug Fixes


src/web/modules/custom/workbc_custom/src/Plugin/Block/ExploreCareersBlock.php
```diff
@@ -111,7 +111,7 @@ public function build() {
 
 
     $options = [];
-    $link = Link::fromTextAndUrl(t($config['label_2'] ?? ""), Url::fromUri('internal:/career-profiles', $options))->toString();
+    $link = Link::fromTextAndUrl(t($config['label_2'] ?? ""), Url::fromUri('internal:/plan-career/explore-careers/career-search', $options))->toString();
```
infojunkie, line 114: This should not be hard-coded. I understand it's not your code, but I am flagging this as something to fix.

zetagraph: Ok, I can create a ticket for it. Who should I assign it to?

infojunkie (commented):
Lgtm, except for the hard-coded paths. Let's agree on a fix for those before merging, and this fix can be done is a subsequent ticket.

zetagraph:
ok, merging as there are a lot of "hot fixes" and moving pieces in here unrelated to the paths. Will create a ticket to update the path separately.

PR #120, opened 2022-10-12
[WR-820] remove hardcoded paths

rjkuyvenhoven:
A-Z Career Profiles page link is now configurable Explore Careers block
Link to "All" for A-Z Career Profiles and Definitions pages are now dynamic.


src/web/themes/custom/workbc/workbc.theme
```diff
@@ -53,6 +53,19 @@ function workbc_preprocess_page(&$variables) {
...
 }
 
+/**
+ * Prepares "view_base_path" variables for unformatted summary view templates
+ * for use with glossary All option.
+ */
+function workbc_preprocess_views_view_summary_unformatted(&$variables) {
+
+  $view = $variables['view'];
+  if ($view->id() == "career_profiles" || $view->id() == "definitions") {
```
infojunkie, line 63: Can we get the path regardless of view name? It does not matter if `getPath` returns nothing since we will only use this variable in specific templates.

infojunkie (commented):
Thanks Ray for picking this up so quickly!

src/web/themes/custom/workbc/workbc.theme
rjkuyvenhoven: view name check has been remove.

infojunkie (approved):
Lgtm, thanks

infojunkie (approved):
Lgtm, thanks

PR #122, opened 2022-10-13
WR-809: bug fixes and cleanup

zetagraph:


src/web/themes/custom/workbc/scss/components/_pagination.scss
```diff
@@ -0,0 +1,32 @@
+.pagination {
```
Mirzero, line 1: Does this still rely on the underlying Drupal styles?
I had named it as an override because it wasn't self-contained styling... it required the Drupal base, and just overrode a few things. Maybe that doesn't matter?

Mirzero (approved):
Other than my one question, looks good.

src/web/themes/custom/workbc/scss/components/_pagination.scss
zetagraph: @Mirzero This is a second pass on the styling as a few things didn't quite match the wires and were raised by QA. Not going for the pixel-perfect here but making a few tweaks and trying to address the visual differences. 

The other issue here is the Bootstrap inheritance. I kept the Bootstrap template for now although we have to keep fighting things coming from bootstrap, not Drupal. The "Drupal Override" convention for these type of components/files didn't quite make sense in this context.

infojunkie (commented):
Lgtm, thanks

PR #123, opened 2022-10-14
[WR-469,470,473]  jobboard search registration login account

anuragparihar1:
This push takes care of most of the Jobboard API-related pages including the Jobsearch section (Listing, search and description pages) from URL **/search-and-prepare-job/find-jobs**, and the Account section (account creation, login, reset the password, Saved Jobs, Recommended Jobs, Job Alert, Personal Settings, and creating user links to Saved Career Profile and Industry Profile pages) from URL **/account**. 
All seems to be working with one minor issue. Currently, dev-api doesn’t allow authentication and throws error 400 (It does the same on dev.workbc.ca site as well), but api-jobboard.workbc.ca does work, but to run it (for testing) two things need to be done
01.	Mention it in settings.local.php line 172 changing from “https://dev-api-jobboard.workbc.ca” to “https://api-jobboard.workbc.ca” for the authentication step & clear cache.
02.	For the main API to work, start your browser with web security disabled. For example “chrome --disable-web-security --user-data-dir=./temp/”


src/web/modules/custom/workbc_jobboard/css/workbc_jobboard.css
```diff
@@ -54,3 +54,18 @@
  width:100%;
  text-align:center;
 }
+app-root .breadcrumbs{display:none;}
```
infojunkie, line 57: I think it would be good to add a large comment here explaining that these styles apply to the embedded Job Board app - or even splitting this into its own CSS file.

src/web/modules/custom/workbc_jobboard/templates/node-page-jobboard-account.html.twig
```diff
@@ -0,0 +1,170 @@
...
+      {{ drupal_region('content_top') }}
+    {% endif %}
+
+    {% if page_body %}
+      {{ page_body }}
+    {% endif %}
+    <div class="layout-container"> 
+    
+    
+    <link rel="stylesheet" href="{{jobboard_api_url}}/bootstrap/bootstrap.min.css?v=Release-1980">
```
infojunkie, line 135: If this section of including Job Board files is repeated across templates, I suggest storing it in a separate template that is included everywhere.

src/web/sites/default/services.yml
```diff
@@ -176,13 +176,13 @@ parameters:
   # for more information about the topic in general.
   # Note: By default the configuration is disabled.
   cors.config:
-    enabled: false
+    enabled: true
```
infojunkie, line 179: I don't think we need to touch any of this config for the Job Board CORS settings.

src/web/sites/default/settings.aws.php
```diff
@@ -22,4 +22,6 @@
 // Single Source of Truth (SSoT) configuration.
 $config['workbc']['ssot_url'] = getenv('SSOT_URL');
 
-$config['jobboard']['api_url'] = getenv('JOBBOARD_API_URL');
+$config['jobboard']['jobboard_api_url'] = 'https://api-jobboard.workbc.ca';
```
infojunkie, line 25: I don't think it's feasible to require testers or other devs to test this code with their browser security disabled. If we want to merge this to `master`, we need to read the API URL from the env var on AWS and solve the actual authentication issue.

src/web/sites/default/settings.local.php
```diff
@@ -238,4 +240,6 @@
 if (class_exists('Kint')) {
   // Set the max_depth to prevent out-of-memory.
   \Kint::$depth_limit = 3;
-}
\ No newline at end of file
+}
+
+header('Access-Control-Allow-Origin: *');
```
infojunkie, line 245: This is a debugging header setting that is not needed.

src/web/sites/development.services.yml
```diff
@@ -8,6 +8,22 @@ parameters:
     debug: true
     auto_reload: true
     cache: false
+
```
infojunkie, line 11: We don't need this section here.

infojunkie (changes requested):
Thanks for this PR! I added some comments that will need changes and let's discuss the authentication issue together.

infojunkie:
Also, @anuragparihar1 you mentioned that you rewrite some URLs inside the Job Board app, but I don't see any code that does that in this PR. Did you miss a JavaScript file?

anuragparihar1:
> Also, @anuragparihar1 you mentioned that you rewrite some URLs inside the Job Board app, but I don't see any code that does that in this PR. Did you miss a JavaScript file?

@infojunkie The URLs are set up in the settings files and used in the twig templates.

src/web/modules/custom/workbc_jobboard/templates/node-page-jobboard-account.html.twig
anuragparihar1: @infojunkie Different file paths are used in both templates (i.e. /dist/jb-account and /dist/jb-search)

src/web/sites/default/services.yml
anuragparihar1: @infojunkie Removing these in services.yml caused the CORS issue to appear again. As of Oct 24, 2022, CORS issue is resolved, but the"'SameSite' attribute set to 'None' without the 'secure' attribute" issue is being experienced. FYI fixing this locally doesn't help, and it shows up on the test2.workbc.ca site as well. 
Keeping the settings for now to avoid the CORS issue

src/web/modules/custom/workbc_jobboard/css/workbc_jobboard.css
anuragparihar1: @infojunkie Block Comment Added

src/web/sites/default/settings.aws.php
anuragparihar1: @infojunkie Updated to Test API path

src/web/sites/default/services.yml
infojunkie: I thought the CORS issue was caused by the service itself. Maybe we need to discuss briefly to remain aligned.

src/web/sites/default/settings.local.php
anuragparihar1: Reverted

infojunkie (approved):
Lgtm, let's merge

PR #128, opened 2022-10-18
WR-741: Labour market base configs

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-741


infojunkie (approved):
Lgtm, thanks

PR #141, opened 2022-10-26
[WR-733] Reports API issue on page when API is unable to connect

anuragparihar1:


src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
```diff
@@ -108,6 +108,19 @@ public function build(){
...
+        }else {
+          return [
+            '#type' => 'markup',
+            '#markup' => 'Explore recent job postings.',
+            '#theme' => 'recent_jobs',
+            '#data' => [],
+            '#sub_title' => $config['job_board_sub_title']??'',
+            '#no_of_records' => $config['job_board_results_to_show']??'',
+            '#readmore_label' => (isset($config['job_board_read_more_button_title'])) ?$config['job_board_read_more_button_title'] : 'View more jobs',
+            '#no_result_text' => 'Error '. $recent_jobs['response'].': Unable to connect to Job Board API.',
```
infojunkie, line 120: It would be best to log the error to the Drupal watchdog instead of the screen. Displaying "unable to connect  to job Board API" is sufficient as far as the user is concerned, but it's the admin who needs a permanent log of the error.

anuragparihar1: Modified

infojunkie (approved):
Lgtm, thanks

PR #145, opened 2022-10-27
Fixes/wr 760  adds calendar hero image

anuragparihar1:


src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
```diff
@@ -108,6 +108,20 @@ public function build(){
             '#noc' => (isset($noc_value)) ? $noc_value : '',
             '#find_job_url'=>\Drupal::config('jobboard')->get('find_job_url'),
           ];
+        }else {
```
infojunkie, line 111: Sory for not flagging this earlier, but you are repeating this `return` statement almost verbatim from above, except for the logging and the empty data. It would be better to:
- if response != 200 then log and keep data empty
- otherwise data = response body
- keep a single return statement

anuragparihar1: @infojunkie Implemented

infojunkie (approved):
Lgtm thanks

PR #147, opened 2022-10-27
Switch aws api

anuragparihar1:
Uses two separate Job Board APIs(i.e. Test & Prod) on AWS to avoid connection issues


infojunkie (approved):
Lgtm, thanks

PR #157, opened 2022-11-03
Job Board configuration updates

infojunkie:
- Rename the frontend and backend job config vars
- Read the frontend config from an env var on AWS


anuragparihar1 (approved):
@infojunkie Thanks for sharing the updated code.

PR #163, opened 2022-11-04
Only job board related external links should get Jobboard related popup

anuragparihar1:


infojunkie (approved):
Thanks for catching this. Lgtm

PR #165, opened 2022-11-07
WR-861: Related topics default image

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-861


infojunkie (commented):
If it has been confirmed that only a single default image is needed for related topics that don't have a hero image, this looks good.

zetagraph:
> If it has been confirmed that only a single default image is needed for related topics that don't have a hero image, this looks good.

As far as I can tell that was confirmed in the ticket:  

> "The same default image, to be provided by AEST, will be displayed in every related topic card if the target page does not have a hero image. "

https://hive.aved.gov.bc.ca/jira/browse/WR-861

infojunkie (approved):
Lgtm, thanks

PR #183, opened 2022-11-16
hero image style, update hero image display; update publication image…

rjkuyvenhoven:
I've added an image style for Hero Images that crops and scales hero images. 

I've applied this image style to all Content Types that display a hero image.

Additionally I've applied the Feature Component style the Image field in the Publications content type.


zetagraph (approved):
LGTM, Thanks @rjkuyvenhoven

PR #185, opened 2022-11-17
WR-992-992 Adds recent jobs section in BC, Region, and Industry profiles

anuragparihar1:


src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
```diff
@@ -85,36 +85,96 @@ public function blockValidate($form, FormStateInterface $form_state) {
...
+      else if($type == 'bc_profile'){
+        $parameters["SearchLocations"] = [
+          [
+            "Region"=> "Cariboo"
+          ],
+          [
+            "Region"=> "Kootenay"
+          ],
+          [
+            "Region"=> "North Coast & Nechako"
```
infojunkie, line 112: Are these names used for UI, or for passing to the Job Board API ?

In the first case, please use "official" spellings for regions:
North Coast and Nechako
Northeast
Cariboo
Kootenay
Thompson-Okanagan
Mainland/Southwest
Vancouver Island/Coast

anuragparihar1: Hello @infojunkie 
These names are passed to API and read how the existing site is passing these.

infojunkie (approved):
Lgtm, thanks. When you deploy this to DEV, please adjust the job board IDs for each industry profile. On my side, I will adjust the content import to set that field correctly going forward.

PR #189, opened 2022-11-18
wr-1079 WorkBC Centres Template and Theme

Mirzero:
This is missing the Related Jobs component as that's not ready yet, but everything else is in place.


infojunkie (approved):
Lgtm, thanks! Looking forward to see those centres themed.

PR #196, opened 2022-11-21
[WR 1105]  Adds the Recent Jobs section to workbc centre pages

anuragparihar1:


src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
```diff
@@ -65,7 +72,7 @@ public function blockSubmit($form, FormStateInterface $form_state) {
     $values = $form_state->getValues();
     $this->configuration['job_board_sub_title'] = $values['job_board_sub_title'];
     $this->configuration['job_board_results_to_show'] = $values['job_board_results_to_show'];
-    $this->configuration['job_board_results_to_show'] = $values['job_board_results_to_show'];
+    $this->configuration['job_board_results_to_show_horizontal_view'] = $values['job_board_results_to_show_horizontal_view'];
```
infojunkie, line 75: Do we need 2 settings for the results? Can't the WorkBC Centres block reuse the same setting as the others?

src/scripts/migration/data/regions_industries.csv
```diff
@@ -1,27 +1,27 @@
...
-"British Columbia","british_columbia","British-Columbia",
-"North Coast and Nechako","north_coast_nechako","North-Coast-Nechako",
-"Northeast","northeast","Northeast",
-"Cariboo","cariboo","Cariboo",
-"Kootenay","kootenay","Kootenay",
-"Thompson-Okanagan","thompson_okanagan","Thompson-Okanagan",
-"Mainland/Southwest","mainland_southwest","Mainland-Southwest",
-"Vancouver Island/Coast","vancouver_island_coast","Vancouver-Island-Coast",
\ No newline at end of file
+"Accommodation and Food Services","Accommodation and Food Services","Accommodation-and-Food-Services","37"
```
infojunkie, line 2: Please review the IDs in light of Meaghan's sheet at WR-1127

src/scripts/migration/data/regions_industries.csv
```diff
@@ -1,27 +1,27 @@
...
+"Manufacturing","Manufacturing","Manufacturing","24"
+"Mining and Oil and Gas Extraction","Mining and Oil and Gas Extraction","Mining-and-Oil-and-Gas-Extraction","21"
+"Professional, Scientific and Technical Services","Professional, Scientific and Technical Services","Professional-Scientific-and-Technical-Services","31"
+"Public Administration","Public Administration","Public-Administration","39"
+"Repair, Personal and Non-Profit Services","Repair, Personal and Non-Profit Services","Repair-Personal-and-Non-Profit-Services","43"
+"Retail Trade","Retail Trade","Retail-Trade","26"
+"Transportation and Warehousing","Transportation and Warehousing","Transportation-and-Warehousing","27"
+"Utilities","Utilities","Utilities","22"
+"Wholesale Trade","Wholesale Trade","Wholesale-Trade","25"
+"British Columbia","british_columbia","British-Columbia","Cariboo,Kootenay,MainlandSouthwest,NorthCoastNechako,Northeast,ThompsonOkanagan,VancouverIslandCoast"
```
infojunkie, line 20: Maybe the list here should be with the same names as the other regions, and gets converted the same way to remove spaces and other chars.

src/scripts/migration/data/regions_industries.csv
```diff
@@ -1,27 +1,27 @@
...
+"North Coast and Nechako","north_coast_nechako","North-Coast-Nechako","NorthCoast&Nechako"
```
infojunkie, line 21: Do you confirm that the expected name is without spaces here?

src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
anuragparihar1: The vertical view shows three jobs by default and the horizontal view displays four.

infojunkie: Yes, but do we ever have both views in the same page? We can set the setting to 3 on one page and 4 on the other.

src/scripts/migration/data/regions_industries.csv
anuragparihar1: fixed

src/scripts/migration/data/regions_industries.csv
anuragparihar1: fixed

src/scripts/migration/data/regions_industries.csv
anuragparihar1: Fixed

src/web/modules/custom/workbc_jobboard/src/Plugin/Block/WorkbcJobboardSidebar.php
infojunkie: Thanks for explaining. I now understamsd that the block is configured only once, hence the need to have two different settings.

infojunkie (approved):
Lgtm, thanks!

PR #235, opened 2022-12-08
[WR-1252] Makes Post-a-Job link configurable

anuragparihar1:


src/web/modules/custom/workbc_jobboard/templates/search-recent-jobs-block.html.twig
```diff
@@ -16,7 +16,7 @@
     <div class="tab-pane fade" id="post" role="tabpanel" aria-labelledby="post-job-tab">
       <h2 class="job-search__title">{{postjob_title}}</h2>
       <div class="job-search__txt">{{postjob_description}}</div>
-      <a class="btn-primary" href="#">Post a Job</a>
+      <a class="btn-primary" href="{{postjob_link_url}}">{{postjob_link_label}}</a>
```
infojunkie, line 19: Assuming this link redirects to an external site (which is safe to assume since there's no such page in our IA) it would be better to use `target="_blank" rel="noopener noreferrer"` to the anchor tag.

infojunkie (changes requested):
Thanks for the PR. Left a comment.

infojunkie (approved):
Lgtm, thanks

PR #239, opened 2022-12-08
WR-844: hero banner updates fixes

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-844

@Mirzero had to reshuffle a few things in order to address the Firefox issue. I've tested things for different hero variations including the Profile ones. If you can review/test as well that would be great. Thank you.


infojunkie (approved):
Lgtm, thanks

Mirzero:
On the Homepage, the stacked image at 920px viewport doesn't look very good... but I'm not sure there's a lot we can do about that and nothing was specced for the no-mans-land between Mobile and Desktop.

Mirzero:
On Career Profiles, at Mobile sizes, the Banner doesn't render correctly. The content box should be the full width/height (there should be no grey background) and the bottom blue border is missing.

image

Mirzero:
BC Profile > Desktop - has double breadcrumb showing
BC Profile > Mobile - has double breadcrumb showing
BC Profile > Mobile - Same Hero Banner issues as Career Profile from previous comment
Regional Profile > Mobile - Same Hero Banner issues as Career Profile from previous comment
Career Profile > Desktop - (Firefox only) Hero Image is full-width instead of right-aligned and constrained

Industry Profile hero banner shows everything correctly at all sizes in Chrome and Firefox.

PR #252, opened 2022-12-15
WR-1262 BC regional profiles interactive map SVG 

anuragparihar1:
with resize below 460px label disappear functionality

This is a self-contained SVG with CSS and JS within it. 
Simply add this using the following code `<div id="bc-map-container" style="text-align:center"><object data="\themes\custom\workbc\assets\images\bc-map.svg" id="bc-map" style="max-width:70%" type="image/svg+xml"></object></div>`
When Window's innerWidth is below 460px, labels would disappear as in the current WorkBC site.


zetagraph (commented):
@anuragparihar1 Looks good to me in general. 

There are a few things though that I was just wondering about. 

- How would this be added to the content/page?

- The current implementation has the `select list` options that make things more accessible for screen readers etc. I guess when things are inside an SVG, that would not be doable.  Maybe these links can be just duplicated in the content in addition to the map itself?

Is there a chance that this can be vulnerable to Cross-Site Scripting attacks via SVG and the JS inside of it? Don't know much about this, just wondering. 

CCing: @rjkuyvenhoven here for an opinion.

anuragparihar1:
> @anuragparihar1 Looks good to me in general.
> 
> There are a few things though that I was just wondering about.
> 
> * How would this be added to the content/page?
> * The current implementation has the `select list` options that make things more accessible for screen readers etc. I guess when things are inside an SVG, that would not be doable.  Maybe these links can be just duplicated in the content in addition to the map itself?
> * Is there a chance that this can be vulnerable to Cross-Site Scripting attacks via SVG and the JS inside of it? Don't know much about this, just wondering.
> 
> CCing: @rjkuyvenhoven here for an opinion.

Hello @zetagraph 

- This image can be added using the code shared in the initial comment. 
- Good catch. The select list will be added to the mobile view when labels disappear
- XSS vulnerability is generally mentioned with the SVG images, but that is limited to user-uploaded images (i.e. we allow website visitors to upload SVGs and they can inject JS within SVG for XSS attack). We are not granting that permission to site visitors, but simply showing our own SVG.

rjkuyvenhoven (approved):
lgtm

PR #263, opened 2022-12-29
WR-1102: Social links breadcrumbs blocks updates

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-1102

- Social and Breadcrumbs block customizations
- Change blocks to use Drupal's page display configurations  

The Breadcrumbs and Social Share Blocks are now on the node level instead of the page level template and can be configured in the Block's visibility settings. This required some reshuffling and the addition of a few views templates.  

There was another ticker floating around in the Backlog regarding these blocks and the inability to configure them in the block's UI settings. This PR should address the issue. 

@infojunkie Please review whenever you have a chance. Thank you.


infojunkie (approved):
Thanks for fixing this. Lgtm

PR #275, opened 2023-01-19
WR-1394: region profile tooltip and label updates

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-1394


infojunkie (approved):
Lgtm, thanks

PR #276, opened 2023-01-20
WR-1354: profile label updates

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-1354


infojunkie (changes requested):
This approach is not what we discussed. We don't want to couple the label of a field to the label of a tab. Instead, I suggest to introduce a new variable `tab` in each tab group that takes the tab label as specified in the wireframe. This variable would be what gets displayed in the tabel label.

infojunkie:
Closing this in favor of https://github.com/bcgov/workbc-main/pull/277

PR #318, opened 2023-03-15
Wr 1503 events not showing

rjkuyvenhoven:
The 5 item limitation that was preventing all events to show on Calendar view has been removed
The Calendar view has been updated to return all events one year old and newer

The Calendar list view has been update to return all events one year old and newer


infojunkie (approved):
Lgtm, thanks

PR #320, opened 2023-03-16
Wr 1503 events not showing

rjkuyvenhoven:
Correct date now displays when event is clicked on the Full Calendar View


infojunkie (approved):
Thanks!

PR #323, opened 2023-03-21
WR-1539: Mobile nav flash on account pages

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-1539

I've added an override to the "<html> top position" which prevents the injected styles from being set. As far as I can tell there are no regressions but things need to be tested.


infojunkie (approved):
Lgtm, thanks

PR #330, opened 2023-04-24
WR-1636: Inline with content media updates

zetagraph:
Tickets: 

https://hive.aved.gov.bc.ca/jira/browse/WR-1636

- Removed File upload from CK Editor
- Updated CK Editor media settings
- Added new Media Display Mode for inline images
- Added new Image styles for CK Editor media embeds


infojunkie:
Thanks Andrei - I tested locally and merged on my machine instead of here. Your code is on DEV now.

PR #334, opened 2023-04-27
[WR-1564, 1564] Media library enhancements

infojunkie:
- WR-1564: Permalink for media library downloads and associated CKEditor button
- WR-1566: Media item add / edit forms patch to support form modes, which allows crop sizes to be selected per field
- WR-1636: Media editor embed styles


infojunkie:
Missing here: Support for form modes in media edit form.

PR #348, opened 2023-06-05
[WR-1603] Centralized resources

infojunkie:
Resource content type, admin Resource view, update Career & Industry form and display settings.


src/config/sync/core.entity_form_display.node.industry_profile.default.yml
```diff
@@ -18,7 +18,6 @@ dependencies:
     - workflows.workflow.editorial
   module:
     - content_moderation
-    - link
```
infojunkie, line 20: @rjkuyvenhoven we cannot remove the old field, because we will lose the data. Instead, we want to deprecate the field.

infojunkie:
@rjkuyvenhoven I would appreciate if you could test the migration on your side. A regular sync will trigger it.

rjkuyvenhoven:
@infojunkie I was able to successfully update and migrate via sync

I did get about 7 "[warning] Message: Could not parse URI entity:node/156" messages

infojunkie:
> I did get about 7 "[warning] Message: Could not parse URI entity:node/156" messages

Thanks @rjkuyvenhoven these are expected - the previous field allowed for internal resources but that's not part of the requirements, so I ignore those during migration.

PR #369, opened 2023-06-21
WR-1560 fix breadcrumb error when editting an archived news article.

rjkuyvenhoven:
Fixed some clumsy code from WR-1322


infojunkie (approved):
Lgtm, thanks

PR #370, opened 2023-06-22
WR-1455 occupational interest filter drop down and tooltip

rjkuyvenhoven:
There are a few styling issues I haven't been able to resolve.

1. The Occupational Interest border is slightly taller than the other fields.  I changed the line height for all the fields. But this one is always slightly taller.
2. The drop down doesn't extend to the full width of the field.
3. The lighter blue focus border doesn't show up for the Occupational Interest field when selecting from the drop down.


src/web/themes/custom/workbc/workbc.theme
```diff
@@ -208,3 +208,30 @@ function workbc_preprocess_menu(&$vars) {
...
   }
 }
+
+
+/**
+ * Implements hook theme_suggestions_paragraph_alter().
+ */
+function workbc_theme_suggestions_form_element_alter(array &$suggestions, array $variables) {
+  
+  if (substr($variables['element']['#id'], 0, 26) == "edit-occupational-interest") {
```
infojunkie, line 218: This seems overly specific, for example against future versions of Views. Can we target a whole id, or something else that is stable in this element?

src/web/themes/custom/workbc/workbc.theme
```diff
@@ -208,3 +208,30 @@ function workbc_preprocess_menu(&$vars) {
...
+
+/**
+ * Implements hook theme_preprocess_form_element().
+ */
+function workbc_preprocess_form_element(&$variables) {
+
+  if (substr($variables['element']['#id'], 0, 26) == "edit-occupational-interest") {
+    $text = \Drupal::config('workbc_custom.settings')->get('tooltips.occupationalinterest');
+    if (is_null($text) || !isset($text)) {
+      $text = 'Careers based on <a href="/plan-career/skills-future-workforce">six interest types</a>.';
```
infojunkie, line 232: Please place this in the template directly - we don't need to parameterize it or even store it as config.

src/web/modules/custom/workbc_custom/src/Form/AdminSettingsForm.php
```diff
@@ -68,6 +68,26 @@ public function buildForm(array $form, FormStateInterface $form_state) {
           '#format' => $text['format'],
         ];
 
+
+        $form['tooltips'] = [
```
infojunkie, line 72: We don't need to parameterize the tooltip. Let's keep it in the theme template.

infojunkie (changes requested):
In general, let's remove the tooltip config and any associated code. We can hard-code this tooltip in the template.

src/web/themes/custom/workbc/workbc.theme
rjkuyvenhoven: I've modified this to use element['#name'] and didn't limit suggestions to 'occupation interest' only

rjkuyvenhoven:
I've removed tooltip configuration and related code.

infojunkie (approved):
Lgtm, thanks. Please don't merge until I test locally and I take a stab at the remaining theming issues.

PR #379, opened 2023-06-29
[WR-1605] resolve issue with differentiating NULL and 0 values

rjkuyvenhoven:


infojunkie (changes requested):
I see similar logic in src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileJobOpeningsByRegionNA.php - please take a look there if anything needs fixing.

rjkuyvenhoven:
Updated Career, BC & Region profiles to differentiate between NULL and 0.

infojunkie (approved):
Lgtm, thanks. Please review each one of these data fields on DEV, as we don't want to introduce regressions with this refactoring. Please focus on N/A vs. 0 values in the different datasets.

PR #381, opened 2023-07-06
[WBCAMS-54] resolve issue with missing icons on FireFox

rjkuyvenhoven:
There seems to be an issue with FireFox not displaying an SVG image the IMG doesn't have a defined width, in this case even if the IMG tag has a WIDTH attribute specified.  Adding a width to the .icon class for IMG tag resolves this issue of the icons not displaying.


infojunkie (changes requested):
Please note that file `src/web/modules/custom/workbc_jobboard/css/workbc_jobboard.css` exists to override Job Board specific CSS. 

We need to either continue using that CSS file, or to move everything in that CSS over to the theme. I prefer the second approach, although I understand it's more work for you.

rjkuyvenhoven:
I've moved the CSS from _`workbc_jobboard.css` to `_job_board.scss` in the WorkBC theme.

infojunkie (approved):
Thanks for cleaning this CSS up. Lgtm now.

rjkuyvenhoven:
FYI, this branch now also includes updates for WEBCAMS-52 and WEBCAMS-53.

PR #383, opened 2023-07-14
[WR-1373] Industry Profile Table Styling

rjkuyvenhoven:
NOTE: The Wage table under Industry Overview tab also didn't match the wireframes, so I updated that table as well.


src/web/themes/custom/workbc/scss/overrides/_industry_profile.scss
```diff
@@ -0,0 +1,17 @@
+
+.industry-profile-table tr {
+  &:nth-child(2n) td {
+    background-color: $workbc_palette_light_primary;
```
infojunkie, line 4: if you are trying to avoid striping, I think that Drupal table theme has a `no_striping`  flag

infojunkie (changes requested):
Please review my comment about table striping

src/web/themes/custom/workbc/scss/overrides/_industry_profile.scss
rjkuyvenhoven: We are not using Drupal to generate those tables, they are built manually in extra fields.

infojunkie: Okay, thanks

infojunkie (approved):

PR #392, opened 2023-07-20
WR-1631 remove decimal points from job openings

rjkuyvenhoven:


src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/BCProfile/BCJobOpeningsForecastChart.php
```diff
@@ -45,9 +45,9 @@ public function viewElements(ContentEntityInterface $entity) {
 
     if (!empty($entity->ssot_data) && isset($entity->ssot_data['regional_labour_market_outlook'])) {
       $data = array();
-      $data[] = floatval($entity->ssot_data['regional_labour_market_outlook']['job_openings_first']);
-      $data[] = floatval($entity->ssot_data['regional_labour_market_outlook']['job_openings_second']);
-      $data[] = floatval($entity->ssot_data['regional_labour_market_outlook']['job_openings_third']);
+      $data[] = intval($entity->ssot_data['regional_labour_market_outlook']['job_openings_first']);
```
infojunkie, line 48: Be careful that if the incoming value has a decimal part, `intval` truncates it instead of rounding it. I made the same mistake recently and the numbers were off by 1 in about half the cases. In this case, `round(floatval())` is needed instead.

rjkuyvenhoven: ok, will update.

infojunkie (approved):
Lgtm now, thanks

PR #393, opened 2023-07-20
WR-1705 add Related Topics Image field, add Related Topics Card content type,…

rjkuyvenhoven:
… update Related Topics Block to handle new image field and content type

### DIFF
--- a/src/config/sync/core.entity_form_display.node.blog.default.yml
+++ b/src/config/sync/core.entity_form_display.node.blog.default.yml
@@ -10,6 +10,7 @@ dependencies:
     - field.field.node.blog.field_image_media
     - field.field.node.blog.field_published_date
     - field.field.node.blog.field_related_topics_blurb
+    - field.field.node.blog.field_related_topics_image
     - field.field.node.blog.field_scheduled_publish
     - field.field.node.blog.field_tags
     - node.type.blog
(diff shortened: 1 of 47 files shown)


src/config/sync/core.entity_form_display.node.blog.default.yml
```diff
@@ -18,7 +19,6 @@ dependencies:
     - content_moderation
     - datetime
     - media_library
-    - media_library_edit
```
infojunkie, line 21: Why is this removed from here and other form displays below? Did we lose the edit icon on media image fields?

src/config/sync/core.entity_form_display.node.related_topics_card.default.yml
```diff
@@ -0,0 +1,122 @@
...
+      placeholder: ''
+    third_party_settings: {  }
+  field_related_topics_image:
+    type: media_library_widget
+    weight: 2
+    region: content
+    settings:
+      media_types: {  }
+      form_mode: media_library
+    third_party_settings: {  }
```
infojunkie, line 59: Make sure to show the media image edit icon here as well.

infojunkie (changes requested):
Looks good! Few comments to answer please.

src/config/sync/core.entity_form_display.node.blog.default.yml
rjkuyvenhoven: Hmmm, I don't know. Will investigate.

infojunkie (approved):
Looks like testing showed the edit icon is still there, so good to go!

PR #399, opened 2023-07-21
WR-1604 job openings   display 0 when negative

rjkuyvenhoven:
I had to update the ssotFormatNA() function to add another variable. 

I think maybe we should consider changing the function so that we pass the value variable and an array of options.  That way we can add new functionality without having a long list of parameters that aren't always used.

ie. ssotFormatNA($value, { 'decimals' => 2, 'suffix' => '%' });


src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileJobOpenings.php
```diff
@@ -44,7 +44,8 @@ public function getLabelDisplay() {
   public function viewElements(ContentEntityInterface $entity) {
 
     if (!empty($entity->ssot_data) && isset($entity->ssot_data['career_provincial']['expected_job_openings_10y'])) {
-      $output = Number_format($entity->ssot_data['career_provincial']['expected_job_openings_10y'],0);
+      $value = $entity->ssot_data['career_provincial']['expected_job_openings_10y'];
+      $output = Number_format($value < 0 ? 0 : $value,0);
```
infojunkie, line 48: we can't use `ssotFormatNA` here?

src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileLabourMarketExpectedOpenings.php
```diff
@@ -44,7 +44,8 @@ public function getLabelDisplay() {
   public function viewElements(ContentEntityInterface $entity) {
 
     if (!empty($entity->ssot_data) && isset($entity->ssot_data['career_provincial']['expected_job_openings_10y'])) {
-      $output = Number_format($entity->ssot_data['career_provincial']['expected_job_openings_10y'],0);
+      $value = $entity->ssot_data['career_provincial']['expected_job_openings_10y'];
+      $output = Number_format($value < 0 ? 0 : $value,0);
```
infojunkie, line 48: we can't use `ssotFormatNA` here?

infojunkie:
I like the idea of adding options instead of parameters. Would this be an opportunity to also merge with `ssotFormatNumber` ?

rjkuyvenhoven:
> I like the idea of adding options instead of parameters. Would this be an opportunity to also merge with `ssotFormatNumber` ?

it might be. will have to give it some thought.

src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileJobOpenings.php
rjkuyvenhoven: Yes, I think we can.

src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileLabourMarketExpectedOpenings.php
rjkuyvenhoven: Yes, I think we can.

src/web/modules/custom/workbc_custom/src/Form/HooOptionsForm.php
```diff
@@ -109,13 +109,12 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
-            //annual wages check
-            $asterisk = '';
-            if (abs($values['wage_rate_median'] - $values['annual_salary_median']) < PHP_FLOAT_EPSILON) {
-              $asterisk = '*';
-            }
-
-            $rows[$key]['wage_rate_median'] = '$'.ssotFormatNumber($values['wage_rate_median'], 2).$asterisk;
+            $options = array(
+              'decimals' => 2,
+              'prefix' => "$$",
```
infojunkie, line 114: I think we're not using HooOptionsForm anymore, but I think there's a single dollar sign here.

src/web/modules/custom/workbc_extra_fields/workbc_extra_fields.module
```diff
@@ -373,39 +373,70 @@ function ssotIndustryName($machine_name) {
...
 
 /*
    custom number format function to add '+' if required
    additional check to set value to 0 if null
    additional special case for zero (no decimals)
+
+   decimal default 0
+   suffix default ""
+   prefix default ""
+   positive TRUE, FALSE default FALSE
```
infojunkie, line 385: positive_sign

src/web/modules/custom/workbc_extra_fields/workbc_extra_fields.module
```diff
@@ -373,39 +373,70 @@ function ssotIndustryName($machine_name) {
...
  */
-function ssotFormatNumber($number, $decimals = 0, $positive = false) {
+function ssotFormatNumber($number, $options = []) {
+  
+  $valid = ['decimals', 'positive_sign', 'suffix', 'prefix', 'na_if_empty', 'no_negative', 'sanity_check'];
+  $error = false;
+  foreach ($options as $key => $option) {
+    if (!in_array($key, $valid)) {
+      \Drupal::logger('workbc_extra_fields')->error("invalid option: @option", ['@option' => $key]);
+      $error = true;
```
infojunkie, line 396: I suggest to log the invalid option kkey but continue displaying the number.

src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileJobOpeningsForecast.php
```diff
@@ -45,9 +45,12 @@ public function viewElements(ContentEntityInterface $entity) {
 
     if (!empty($entity->ssot_data) && isset($entity->ssot_data['career_provincial'])) {
       $data = array();
-      $data[] = intval($entity->ssot_data['career_provincial']['job_openings_first']);
-      $data[] = intval($entity->ssot_data['career_provincial']['job_openings_second']);
-      $data[] = intval($entity->ssot_data['career_provincial']['job_openings_third']);
+      $value = intval($entity->ssot_data['career_provincial']['job_openings_first']);
+      $data[] = $value < 0 ? 0 : $value;
```
infojunkie, line 49: isn't that `no_negative`  ?

infojunkie (changes requested):

src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileJobOpeningsForecast.php
rjkuyvenhoven: This is numeric data used for generating a chart, ssotFormatNumber() returns a string value.

src/web/modules/custom/workbc_extra_fields/workbc_extra_fields.module
rjkuyvenhoven: documentation updated

src/web/modules/custom/workbc_extra_fields/workbc_extra_fields.module
rjkuyvenhoven: modified to display number even if invalid option found.

infojunkie (approved):
Lgtm, thanks

PR #400, opened 2023-07-24
[WR-1686] give appropriate users permission to use media_file_delete module functionality.

rjkuyvenhoven:


infojunkie (approved):
Lgtm, thanks

PR #406, opened 2023-07-26
[WR-1345] fix BC profile header font size

jonathan-longe:
Increase header font size to 45px as requested

Screenshot from 2023-07-25 17-23-20


infojunkie (approved):
Lgtm, thanks

PR #408, opened 2023-07-26
[WR-1332] display job titles in 2 columns

jonathan-longe:
Screenshot from 2023-07-26 14-17-28


infojunkie (approved):
Lgtm, thanks. Note that master branch is currently locked so nothing further to do until it is unlocked.

PR #409, opened 2023-07-27
[WR-1332] revision 2 - fix more link

jonathan-longe:
Changed the greater than sign after the "View more" text (see below):
Screenshot from 2023-07-27 09-35-05

The previous version looked like this:
Screenshot from 2023-07-26 14-17-28

Below is the image showing what the designer requested. 
WR-1332 Wireframe

Also, this PR modifies the behavior of the "View more >" link.  Now, when users hover over the link, an underline appears.  I think this is consistent with other links on the site, but I'm open to alternatives if that's not appropriate.


infojunkie (approved):
Thanks, lgtm. Since there is a difference between angle brackets, please make sure that the angle bracket used is the same as on different cards of the live site.

jonathan-longe:
> Thanks, lgtm. Since there is a difference between angle brackets, please make sure that the angle bracket used is the same as on different cards of the live site.

Excellent tip @infojunkie I've replaced the ">" with the CSS chevron that's used elsewhere on the site.

jonathan-longe:
Here's what the CSS chevron looks like:

Screenshot from 2023-07-27 10-19-27

PR #410, opened 2023-07-27
[WR-1332] revision 3

jonathan-longe:
Apologies colleagues. 3 PR's for a simple front-end fix! 

This PR fixes a bug spotted in DEV where the chevron was wasn't being hidden after the "View more" link was clicked.


infojunkie (changes requested):
Missing here:
- Add pointer cursor to "View more" link
- Add a "View less" link with the same styling as "View more" and that is activated by a boolean parameter

src/web/themes/custom/workbc/js/static-load-more.js
```diff
@@ -4,27 +4,54 @@
   let initialize = function (containerJquery) {
     let container = $(containerJquery);
     let initialCount = container.data('static-load-more-initial');
+    let moreText = container.data('static-more-text');
```
infojunkie, line 7: Previously, "View more" was hard-coded in the template. Now that it became a parameter, what happens to all previous locations that do not declare that parameter?

src/web/themes/custom/workbc/templates/content/node--career-profile--full.html.twig
```diff
@@ -392,16 +392,20 @@
             <div class="profile-content-section" id="job-titles">
               <h2 class="profile-content-section__header">{{ viewmodel['tab_sections']['career_overview']['job_titles_title'] }}</h2>
 
-              <div data-static-load-more-container data-static-load-more-initial="6" data-static-load-more-step="200">
+              <div data-static-load-more-container
+                   data-static-load-more-initial="6"
+                   data-static-load-more-step="200"
+                   data-static-more-text="View more"
+                   data-static-less-text="View less">
```
jonathan-longe, line 399: Added two parameters to `static-load-more.js`.  Now, when the 'data-static-less-text' is populated a show-less link will appear in the expanded view.

src/web/themes/custom/workbc/js/static-load-more.js
```diff
@@ -4,27 +4,54 @@
...
     hiddenItems.slice(0, stepCount).show();
-      
-    if(hiddenItems.length <= stepCount) {
-      let trigger = container.find('[data-static-load-more-trigger]').first();
+    let trigger = container.find('[data-static-load-more-trigger]').first();
+
+    // if lessText is not set, hide the 'show less' link
+    if(hiddenItems.length <= stepCount && ! lessText) {
       trigger.hide();
     }
```
jonathan-longe, line 35: If the `data-static-less-text` parameter is not populated, the less-text link is hidden

src/web/themes/custom/workbc/js/static-load-more.js
```diff
@@ -4,27 +4,54 @@
...
 
     items.slice(initialCount).hide();
 
     trigger.on('click', function() { loadMore(container) });
 
+    // if moreText parameter is populated, update link text
+    if(moreText) {
+      trigger.text(moreText)
+    }
+
```
jonathan-longe, line 19: The changes made to `static-load-more.js` and backward compatible.  They do not disrupt the existing users of the functionality.

src/web/themes/custom/workbc/scss/components/_career-profiles-search.scss
```diff
@@ -28,16 +28,13 @@
   list-style-position: inside;
 }
 
+#job-titles .pointer:hover {
+  cursor: pointer;
+}
+
```
jonathan-longe, line 34: Unfortunately, I had to add additional CSS to change the cursor to a pointer on hover.  The existing CSS only works on anchor tags that include an `href=''`.  Since the `<a>` tags to "View more" and "View less" cannot include an `href`, I've added this additional CSS.

src/web/themes/custom/workbc/templates/content/node--career-profile--full.html.twig
```diff
@@ -392,16 +392,20 @@
...
                 {% for row in viewmodel['tab_sections']['career_overview']['job_titles_list'] %}
                   {% for jobtitle in row %}
                     <div class="job-title-item">{{ jobtitle.value }}</div>
                   {% endfor %}
                 {% endfor %}
                 </div>
-                <div class="static-load-more-button-container">
-                  <a data-static-load-more-trigger>View more</a>
+                <div class="static-load-more-button-container megamenu-splash__actions">
+                  <a class="action-link pointer" data-static-load-more-trigger></a>
```
infojunkie, line 408: what happens if `data-static-more-text` is not set?

jonathan-longe: Good question @infojunkie , `data-static-more-text` is only used in conjunction with 'data-static-less-text'.  In the case where `data-static-more-text` is not set, the text from the anchor tag is used by default.  For example, `<a ... >View more</a>`

src/web/themes/custom/workbc/js/static-load-more.js
jonathan-longe: Where the parameter is not set, we default to the text in the anchor tags, like this, `<a...>View more<a>`.  The parameter is only required to use the show-less functionality.

infojunkie (approved):
Lgtm now, thanks! Please take your time to verify everything on DEV.

PR #413, opened 2023-07-31
[WR-1357] fix employment rate definition

jonathan-longe:
This PR fixes the definition of "Employment Rate".  I had inadvertently copied the definition of "Unemployment Rate".
Since this is a simple text change, I don't believe there are any build artifacts to commit with this PR.


infojunkie (approved):
Lgtm, thanks

PR #414, opened 2023-08-01
[WR-1458] add responsiveness to employment-by-age-gender table

jonathan-longe:
Client requested:

WR-1458 wireframes

Here's what the block looks like when viewed at a screen width less than 600px.

Screenshot from 2023-08-01 09-51-42

And here's what the block looks like when viewed on a large screen:

Screenshot from 2023-08-01 09-54-28


infojunkie (approved):
Beautiful! Thanks, lgtm.

PR #415, opened 2023-08-02
[WR-1326] fix decapitation of image

jonathan-longe:
### Do not merge until deployment activities are complete

The client provided the following two screenshots and asked that they be fixed.

WR-1326 firefox hero
WR-1326 hero ipad

The page above now looks like this:
Screenshot 2023-08-02 094017

The current template system treats all "hero" images the same (expect the image on the home page).  After this PR is deployed, it's not clear if the presentation of other hero images will be acceptable.

For example, in order to have the image cover most of the width of a wide-screen page, the lower portion of the image is automatically cropped like this:

Screenshot 2023-08-02 094600

The alternative -- which I think doesn't look as nice -- is to always display the entire image (see below).  This leaves a large amount of whitespace in the middle of the page.

Screenshot 2023-08-02 094851


jonathan-longe:
As part of this same ticket. the client has provided the following screenshot of the home page, but it's not clear to me how the presentation should be changed.

WR-1326 ipad home

jonathan-longe:
@infojunkie when you have sec, can you take another look at this PR?

src/web/themes/custom/workbc/scss/components/_hero-banner.scss
```diff
@@ -20,7 +20,8 @@
     bottom: 0;
     right: 0;
     z-index: 0;
-    max-height: none;
+    height: 397px;
```
infojunkie, line 23: It's not clear to me where these numbers are coming from. If they are copied from someplace else, would it make sense to have them as variables instead?

jonathan-longe: That's an excellent idea @infojunkie !  Standby fix in the works ...

infojunkie (approved):
Beautiful, thanks!

infojunkie:
But please wait before you merge back into master we're not in the clear yet.

PR #418, opened 2023-08-04
WR-1349 interactive map

rjkuyvenhoven:
Looking for a review of current state as well as some help on a couple issues I can't seem to figure out.   

1. Labour Market Monthly label lines aren't working as they are with the other map instances (they should all be using the same CSS)
2.  Highlight background color for table row on selected region doesn't work on alternate rows. ( 

Career Profile - Labour Market Statistics tab (2 maps)
http://workbc.docker.localhost:8000/career-profiles/registered-nurses-and-registered-psychiatric-nurses

BC Profile - Labour Market Statistics tab
http://workbc.docker.localhost:8000/region-profile/british-columbia

Labour Market Monthly
http://workbc.docker.localhost:8000/research-labour-market/bcs-economy/labour-market-monthly-update

Industry Profile
http://workbc.docker.localhost:8000/industry-profile/accommodation-and-food-services


src/web/modules/custom/workbc_extra_fields/js/interactive-map.js
```diff
@@ -0,0 +1,35 @@
+let regions = ['british_columbia', 'cariboo', 'kootenay', 'mainland_southwest', 'north_coast_nechako', 'northeast','thompson_okanagan', 'vancouver_island_coast'];
```
infojunkie, line 1: This file needs a standard Drupal preamble - refer to `workbc_joboard.js` for an example.

At the very least:
```
(function ($, Drupal, once) {
	Drupal.behaviors.interactiveMap = {
    attach: function (context, settings){
      once('interactiveMap', 'your-selector', context).forEach(function() {
...
```

src/web/modules/custom/workbc_extra_fields/js/interactive-map.js
```diff
@@ -0,0 +1,35 @@
...
+let currentRegion = [0,0,0,0];  // 4 maps
+
+
+function mapHoverOn(map, region) {
+
+  var element = document.querySelector("#workbc-interactive-map-" + map + " #interactive-map-" + regions[region]);
```
infojunkie, line 7: I don't think each region needs to be an id. A class is enough to target it.

src/web/modules/custom/workbc_extra_fields/src/Plugin/ExtraField/Display/CareerProfile/CareerProfileJobOpeningsByRegion.php
```diff
@@ -92,17 +92,25 @@ public function viewElements(ContentEntityInterface $entity) {
       $regions[] = $region;
     }
 
-    $module_handler = \Drupal::service('module_handler');
-    $module_path = $module_handler->getModule('workbc_extra_fields')->getPath();
+    // $module_handler = \Drupal::service('module_handler');
```
infojunkie, line 95: If no longer needed, you can delete these lines.

src/web/modules/custom/workbc_extra_fields/js/interactive-map.js
rjkuyvenhoven: we need to target a specific region.  The id uniquely identifies the region.

infojunkie: According to the spec, IDs need to be unique in the whole document https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

My understanding is that if we have 2 maps, the region id above would be repeated. If that's correct, then I suggest to make this a class instead, since those regions are essentially static.

rjkuyvenhoven: modified to use class instead of id

infojunkie:
Some comments after testing this PR locally:

- The map is not centere within its container card
Screenshot from 2023-08-04 16-18-41

- The browser should scroll to the table when a region is selected (e.g. use [`scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView))

- The highlighted row is not using the correct styles as per wireframes http://aved.number41media.com/Wireframes/design-refresh/Beta2022/#id=igtdif&p=career_profile_-_approved&g=1

- Some highlighted rows are not highlighted at all (possibly due to striping, please check the specificity of the CSS rules)
Screenshot from 2023-08-04 16-19-11

- The B.C. average row is not styled correctly 
screenshot-workbc docker localhost_8000-2023 08 04-16_23_25

rjkuyvenhoven:
Interactive Maps have  been completed
Refactoring the extra field tables associated with the Interactive Maps as been completed

infojunkie (approved):
I tested locally and the maps work great! I made a couple of minor changes. Thanks for that great PR.

PR #420, opened 2023-08-08
[WR-1483] centre table header on mobile

jonathan-longe:
The client provided the following screenshot showing a discrepancy between the wireframes and the actual website when viewed using a mobile browser.

alignment spacing

The issue was fixed by centering the header as shown below.

Screenshot from 2023-08-08 10-25-35

Note: I've centred the header for all similar tables -- not just this table as this appears to be the wish of the themers.


infojunkie (approved):
Lgtm, thanks

PR #422, opened 2023-08-08
[WR-1535] make high opportunity table responsive

jonathan-longe:
The client uploaded the following image showing a discrepancy between the wireframes and production:

Mobile HOO table

Below is the fix as shown on iPhone:

Screenshot from 2023-08-08 15-10-35


src/web/themes/custom/workbc/templates/views/views-view-table.html.twig
```diff
@@ -100,7 +100,7 @@
               {% set column_classes = column_classes|merge(['views-field-' ~ field]) %}
             {% endfor %}
           {% endif %}
-          <td{{ column.attributes.addClass(column_classes) }}>
+          <td{{ column.attributes.addClass(column_classes) }} data-label="{{ header[key].content }}">
```
jonathan-longe, line 103: @infojunkie I think this is safe to apply to all tables displayed using `views-view-table.html.twig` but I'd appreciate your thoughts

jonathan-longe:
Note: in the iPhone screenshot above, the field content "Diploma/Certificate" is not wrapped as requested in the wireframe.  Would it be possible to change the content to read "Diploma / Certificate" instead so it wraps as requested?

src/web/themes/custom/workbc/templates/views/views-view-table.html.twig
infojunkie: Do you know if `header[key]` is always present ?

infojunkie:
> Note: in the iPhone screenshot above, the field content "Diploma/Certificate" is not wrapped as requested in the wireframe. Would it be possible to change the content to read "Diploma / Certificate" instead so it wraps as requested?

The string "Diploma/Certificate" is part of government data via SSoT. It does not reside in the CMS. I will bring this up in the ticket.

src/web/themes/custom/workbc/templates/views/views-view-table.html.twig
jonathan-longe: The preamble of the twig file suggests the header is always present, but I'm not willing to risk it.  I've added a safety.  @infojunkie let me know if you think that's sufficient.

PR #428, opened 2023-08-09
[WR-1479] fix-spacing-above-labour-market-content-source

jonathan-longe:
The client provided the following screenshot showing insufficient space above the "Source ...".  The lack of space is particularly noticeable when viewed on a mobile device.

spacing before source

Here's how the fix looks on an iPhone.  

Screenshot from 2023-08-09 13-31-28


infojunkie (approved):
Lgtm, thanks!

PR #432, opened 2023-08-10
[WR-1326] always-show-full-home-page-hero-image

jonathan-longe:
Change the breakpoint so more of the hero image on the home page is shown.  Instead of obscuring a portion of the hero image, the client has requested that the hero image be stacked.

Now, when the home page is viewed at a screen width more than 1300px, it appears like this:
(Note the whole hero image is shown)

Screenshot from 2023-08-10 17-29-01

But, when the home page is viewed at a screen widh less than 1300px, it stacks the hero image above the "Find a job" form, so at no point is the hero image obscured by the form.

Screenshot from 2023-08-10 17-32-10


infojunkie (approved):
Lgtm, thanks

PR #435, opened 2023-08-12
[WR-1719] theme-bc-map-and-table-for-mobile-devices

jonathan-longe:
When the interactive BC maps are shown on a screen wider than 600px, we display labels as shown below:
Screenshot from 2023-08-14 10-19-51

However, when the screen width is less than 600px, we remove the labels as shown below:

Screenshot from 2023-08-14 10-15-48

When users click on a region of the map, the area is highlighed in light blue and the user is shown the statistics for the area in the table below.

Screenshot from 2023-08-14 10-16-04

Notes: 
 - This PR includes the themeing of the interactive map for mobile of the "Job Openings by Region" and the "Employment by Region" tickets WR-1719 and ~~WR-1721~~ WR-1718
 - The mobile wireframes show that only the stats for the region selected are shown, but the filtering logic was not implemented due to complexity (see notes in the respective parent tickets).  We may want to consider highlighting the selected table row like we do on screen's wider than 600px but this would require a change to the wireframes.
 - On moblie devices, the BC map slightly is slightly larger than the screen size.  It looks like the map has a hard coded width of 400px and the click targets are based on the hard coded map dimensions.  I'm happy to address this and make the map size scale if the map appears too wide.  Please let me know if you think this work is necessary.


infojunkie:
Thanks @jonathan-longe as per discussion, we want to add the following to this PR:
- Resizing the map to fix within device screen width
- Highlight the selected region as per the B.C. Average row at http://aved.number41media.com/Wireframes/design-refresh/Mobile2022/#id=zhgtdj&p=l3_-_bc_profile_-_approved&g=1 (Labour Market Statistics tab)

jonathan-longe:
This PR has been updated so the map is resized correctly and the selected region is highlighted in the table. 
Screenshot from 2023-08-15 11-23-06

jonathan-longe:
Another image showing the other BC map on the same page

Screenshot from 2023-08-15 11-28-23

rjkuyvenhoven:
lgtm, thanks

PR #440, opened 2023-08-16
WR1677 file deletion

rjkuyvenhoven:
The purpose of this PR is to ensure unused media items are deleted from the site.

Update site configuration:
1. set Revision checkbox default to UNCHECKED for media items
2. set Drupal to mark unused managed files to TEMPORARY

Media add/edit form alter
1. hides Revision Information tab

Post Update
1. post update function to delete existing non-current revisions for media items

As a result of these changes, current and future unused managed files will be marked as TEMPORARY and will be deleted from the site after 6 hour via CRON


jonathan-longe (approved):
Thanks for the opportunity to review.  Lots of learning for me here!

PR #444, opened 2023-08-17
[WR 1538] fix side nav menu broken on mobile

jonathan-longe:
This PR fixes the broken mobile menu.  Previously, when the menu was activated the content was pushed right leaving a view content width that was wider than view port.  This caused major problems.

### Previous
Screenshot from 2023-08-17 10-11-47

Now, the mobile menu is written on top of the primary content as shown below:

Screenshot from 2023-08-17 12-59-00

If the user activates the menu after scrolling down the page, the menu is displayed from the top of the current view port.

Screenshot from 2023-08-17 12-59-24

## Important

In compliance with guidance from the mobile menu (mmenu) library, the following configuration changes should be unchecked on:
`/admin/config/user-interface/responsive-menu`:

Screenshot from 2023-08-17 13-08-25

Screenshot from 2023-08-17 13-08-55

@rjkuyvenhoven can you provide advice on how changes like these should be scripted for deployment?


src/web/themes/custom/workbc/scss/layout/_responsive-menu.scss
```diff
@@ -1,3 +1,226 @@
...
+.mm-panel a:hover {
+  border: none;
+}
+
+ul.menu.mm-listview {
+  padding-left: 0;
+  margin-left: 0;
+}
+
+// CSS above this point is provided by Drupal module "drupal/responsive_menu"
```
jonathan-longe, line 221: Rather than loading CSS from the mmenu library, I'm including it here in source control.  This seems safer and prevent changes to the library CSS changing our site.

src/web/themes/custom/workbc/templates/layout/html.html.twig
```diff
@@ -44,16 +44,18 @@
     <js-placeholder token="{{ placeholder_token }}">
   </head>
   <body{{ attributes.addClass(body_classes) }}>
-    {#
+    <div id="page">
+      {#
```
jonathan-longe, line 48: The mobile menu requires that our content be wrapped in `<div id="page">`.  The Drupal library can automatically add a wrapper, but this can cause unnecessary page reloads.

src/web/themes/custom/workbc/scss/layout/_responsive-menu.scss
rjkuyvenhoven: you are just copying over the CSS and not making any changes?

jonathan-longe: Yes, everything above line 221 is copied from the responsive_menu library.

src/web/themes/custom/workbc/templates/layout/page.html.twig
```diff
@@ -54,7 +54,7 @@
 {# Main Nav #}
 {% include '@workbc/includes/main-nav.html.twig' %}
 
-<main role="main" class="page-wrapper">
+<main role="main">
```
jonathan-longe, line 57: Just realized that this change is not required.  I'm going to restore the file from the master branch.

src/web/themes/custom/workbc/scss/layout/_responsive-menu.scss
rjkuyvenhoven: In theory, we have control over when/if the mmenu library is updated and can test any updates to see if they have a negative impact on the site and revert to previous version if necessary. Therefore, I think it best to load the CSS from the library.

jonathan-longe: I'm not opposed to reverting to the way we had it previously, but the module maintainer suggests the following:
> It should also be noted that the default css that comes with the module provides
> some very basic styling and should be copied and pasted into your theme's css
> so that you can modify it to fit your theme's style. Once you've copied and
> pasted the css into your stylesheet you can disable the inclusion of the
> module's css on the settings page.

See: https://git.drupalcode.org/project/responsive_menu/-/blob/4.3.x/README.md

rjkuyvenhoven: ok, let's go with that.

rjkuyvenhoven (approved):
lgtm, thanks!

PR #445, opened 2023-08-17
[WR-1393] fix presentation of regional profile employment stats

jonathan-longe:
This PR fixes the following:

On the BC Profile -> LMS tab
See:  http://workbc.docker.localhost:8000/region-profile/british-columbia

- [x] prevent premature wrapping of the word "average" when rendered on mobile devices

### Before

Screenshot from 2023-08-17 16-18-33

### After

Screenshot from 2023-08-17 16-20-12

On the Regional Profile pages -> LMS tab
See: http://workbc.docker.localhost:8000/region-profile/thompson-okanagan

- [x] prevent premature wrapping of the following tables when rendered on mobile devices.
- [x] remove the text, "(average)" from the regional stats table -- but not the BC table

### Before 

Screenshot from 2023-08-17 16-24-22

### After 

Screenshot from 2023-08-17 16-22-23


rjkuyvenhoven (approved):
lgtm, thanks

PR #446, opened 2023-08-18
WR-1439 graham island hot spot

rjkuyvenhoven:
added new map hot spot for Graham Island


jonathan-longe (approved):
Looks great!  Thanks for the opportunity to review.

PR #447, opened 2023-08-18
[WR-1073] display "On this page" block under title on mobile

jonathan-longe:
When pages without a "hero" image are displayed on small screen devices, the "On this page" block should appear below the page title.  On larger screen devices, the "On this page" block should appear in the left sidebar as it does currently.   Like this:

Screenshot from 2023-08-18 11-32-45

Screenshot from 2023-08-18 11-33-21

Screenshot from 2023-08-18 11-33-39


src/web/themes/custom/workbc/templates/content/node--page--full.html.twig
```diff
@@ -138,6 +138,12 @@
       {{ drupal_region('content_top') }}
     {% endif %}
 
+    {% if left_block|striptags|trim %}
+      <div class="on-this-page">
+        {{ left_block }}
+      </div>
+    {% endif %}
```
jonathan-longe, line 145: The `{{ left_block }}` is now duplicated in `node--page--full.html.twig` and then the blocks are toggled on / off using CSS depending on the screen width.

rjkuyvenhoven (approved):
lgtm, thanks!

PR #449, opened 2023-08-25
[WR-1538] - fix mobile menu for iOS v15

jonathan-longe:
Tiny change to fix previously merged PR #444 that caused issues clicking buttons and links on iOS version 15.2 but appears to have no issues on other iOS versions and other mobile devices.

This PR yields no visual changes so no screenshots are included here.


src/web/themes/custom/workbc/scss/layout/_responsive-menu.scss
```diff
@@ -323,12 +323,11 @@ html {
   z-index: 99;
 }
 
-a#mm-0 {
+body.mm-wrapper--opened a#mm-0 {
```
jonathan-longe, line 326: Safari version < 15.5 does not recognise the html "inert" attribute, so this is the workaround.

rjkuyvenhoven (approved):
lgtm, thanks

PR #452, opened 2023-09-05
[WR-1538] do not load boilerplate mmenu css

jonathan-longe:
Change Drupal configuration using `drush config:export` so the mobile menu's (mmenu) boilerplate CSS file are not imported but instead rely on the customized one we wrote.  This change aligns with the guidance mmenu suggests on their configuration page.


infojunkie (approved):
Lgtm, thanks

PR #453, opened 2023-09-06
[WR-1386] update job board commit hash

jonathan-longe:
Update job board commit hash in `settings.php`


infojunkie (approved):
That's exactly it, thanks!

PR #454, opened 2023-09-07
[WR-1703] Fix duplicate titles in sidebar navigation

infojunkie:
Placing a second version of the sidenav menu inside the article body caused 3 bugs:
- The h2 "On this page" was picked up as a title to be added to the menu
- Each subsequent h2 was added twice to the menu
- Each anchor was added twice to its corresponding h2 

This PR solves these bugs.


jonathan-longe (approved):
Thanks for the opportunity to review.  LGTM!

PR #455, opened 2023-09-13
[WR-1470] fix presentation of high opportunity occupation table

jonathan-longe:
# Do not merge until after the release

image

Fixes to the HOO table shown above include:

- Occupation column is slightly wider
- Education Requirements column is now centre-aligned and the column width reduced
- Hourly Wage column has been renamed "Median Hourly Earnings" and the header and data is now right justified
- Job Openings column is right justified and the column has been renamed "Job Openings to 2032"
- Occupational Interest column is now centre-aligned


src/web/modules/custom/workbc_custom/workbc_custom.module
```diff
@@ -101,7 +101,11 @@ function workbc_custom_views_pre_render(ViewExecutable $view) {
   if ($view->id() == 'high_opportunity_occupations') {
     $schema = ssotSchema("");
     $range = ssotParseDateRange($schema, 'high_opportunity_occupations', 'openings_forecast');
-    $view->field['openings_forecast']->options['label'] = "Job Openings to " . $range;
+    if (preg_match('/(\d{4})$/', $range, $end_of_range)) {
```
infojunkie, line 104: what happens if we get a single year instead of a year range here?

jonathan-longe: It should handle that gracefully.  I tested the regex expression with both and when a 4 digit year is supplied it returns the last 4 digits of the string.

infojunkie: sounds good, thanks!

infojunkie (approved):
Lgtm, thanks

PR #456, opened 2023-09-13
WR-1715 Prevent content editor from selecting different Content Group

rjkuyvenhoven:
hide Content Group for users who don't have permission to bypass content groups, set Content Group field value to User's content group when adding new content.


infojunkie (approved):
Lgtm, thanks

infojunkie:
@rjkuyvenhoven can you please relabel this PR with the ticket number?

PR #459, opened 2023-09-13
WR-1608 google translate english issue

rjkuyvenhoven:
New patch for gtranslate contrib module replacing [select_default_value.patch] - removes Select Language option from drop down and removes highlighting of default language in drop down options list


infojunkie (approved):
Lgtm, thanks

PR #460, opened 2023-09-13
[WR-1427] fix career profile styling

jonathan-longe:
# Do not merge until after the release

image

As per  ticket instructions, increase font size and space above the subheadings in the career profile labour market statistics tab. Note: the blue separator line was already present underneath the subheadings.


infojunkie (approved):
Lgtm, thanks

PR #461, opened 2023-09-13
[WR-1399] fix recent jobs styling

jonathan-longe:
# Do not merge until after release

image

From the ticket:

> Please correct the following to match the wireframes:
> 
> Text "Explore recent job postings" font should be 16
> Within the 'Recent Jobs' component:
>
> The employer font should be 16
> The location font should be 15
> There is too much space after the colon in the posted text


infojunkie (approved):
Lgtm, thanks

PR #462, opened 2023-09-13
WR-1196 definitions letter highlighting issue

rjkuyvenhoven:
The same issue existed on the Career Profiles Search A-Z page, which has also been resolved in this PR.


infojunkie (approved):
Thanks for fixing in both places! Lgtm

PR #464, opened 2023-09-13
[WR-1423] fix labour market statistics table styling

jonathan-longe:
# Do not merge until release completed

Fix two tables as per ticket instructions:

image

image

From the ticket:

> As per Wireframe and Confluence, please update:
> 
> - Employment by Region table > Column heading - % Employment TO: % Employment of this Occupation, ensuring it follows correct 'wrap text' (displaying on two lines)
> -  Job Openings by Region table > Column Size - should have equal column widths (example Employment by Region table)
> -  Job Openings by Region table > Table heading- Avg Annual Employment Growth should __ be text wrapped, displaying on two lines (this should occur once column size is reduced).
>    - Avg should be displayed with a period after (Avg.) - as per confluence (image attached) 
> - Expected Job Openings Data Card should read: Forecasted Job Openings. (image attached)
> - For both tables within LMS tab:
>   - Table within shadow box is too narrow compared to wireframes, too much white space on sides
>   - Rows are too high compared to wireframe
>   - Numbers should be right aligned, not left (just numbers)
>     - when numbers are right aligned, the column headers should be too


jonathan-longe:
@infojunkie we received an 'okay' from the client to allow column title text to wrap naturally.  The client will consider further revisions in the future if it doesn't look right.

infojunkie (approved):
Lgtm, thanks

PR #467, opened 2023-09-14
WR-1380 - Employment by Region table column display incorrect

rjkuyvenhoven:
add year to value columns, right align header and data in value columns


infojunkie (approved):
Lgtm, thanks

PR #470, opened 2023-09-14
[WR-1459] make labour market monthly update map responsive

jonathan-longe:
The map displays as shown when the device view is less than 600px wide:

image

The map displays as shown when the device view is greater than 600px wide:

Screenshot from 2023-09-14 12-15-14


infojunkie (approved):
Lgtm, thanks! Let's merge this PR and verify on DEV.

PR #473, opened 2023-09-18
WR-1487 Search results displaying incorrectly

rjkuyvenhoven:


infojunkie (approved):
Great solution, thanks!

PR #474, opened 2023-09-18
WR 1706 Interactive map component

rjkuyvenhoven:


infojunkie:
I did the following changes to the component:
- Change the paragraph type description text
- Change each link field to: required value, no link text, only internal links

With these changes, I was able to create a map section and click the region links. I also tested editing a map, and adding 2 maps on the same page. All working well!

rjkuyvenhoven:
updated permissions so all users will be able to view Interactive Map component.

will add deployment instructions to ticket.

PR #477, opened 2023-09-19
[WR-1070] show 3 browse related topics in landscape

jonathan-longe:
# Do not merge until the deployment is completed

When in portrait mode, the two outside related topics are shown only partially:

Screenshot from 2023-09-18 18-00-23

But now, in landscape, the two outside related topics are shown in full:

Screenshot from 2023-09-18 18-01-04


infojunkie (approved):
Lgtm, thanks

src/web/modules/custom/workbc_custom/templates/related_topics_block.html.twig
jonathan-longe, line 1: This file appears not to be used.  Please confirm it's okay to delete.

infojunkie: Thanks for double-checking. I would be surprised that it's not used. It's being referenced by the plugin `RelatedTopicsBlock` and elsewhere.

jonathan-longe: Thanks, I'll restore it.

PR #478, opened 2023-09-20
WR-1361 half width icon cards

rjkuyvenhoven:


infojunkie:
Thanks @rjkuyvenhoven I will download and test this locally as well.

infojunkie:
Works well for me. One question to @jonathan-longe: The cards without bottom links look the attached. Is there any explicit expectation about this case (e.g. to center the icon / text) ? If not indicated in the ticket, no need to bring it up.

screenshot-workbc docker localhost_8000-2023 09 20-12_22_52

jonathan-longe:
> Works well for me. One question to @jonathan-longe: The cards without bottom links look the attached. Is there any explicit expectation about this case (e.g. to center the icon / text) ? If not indicated in the ticket, no need to bring it up.

@infojunkie good question! I'm not aware of such a requirement.  Every example in the wireframes includes a link in the footer.  I'd suggest we leave it as is and adjust in the future if necessary.

infojunkie:
Sounds good, thanks @jonathan-longe

PR #479, opened 2023-09-20
[WR-948] reduce video thumbnail size

jonathan-longe:
# Do not merge until production deployment is complete

Below is a screenshot showing how the 2-column thumbnails appear:

image


infojunkie:
Testing this: would it be possible to avoid the scrolling in the popup window? I realize it also exists in the Video Library popup (which is a good thing for consistency). Can @jonathan-longe advise on this?
Screenshot from 2023-09-20 11-37-49

jonathan-longe:
> Testing this: would it be possible to avoid the scrolling in the popup window? I realize it also exists in the Video Library popup (which is a good thing for consistency). Can @jonathan-longe advise on this?

Well spotted.  I think we should remove it using CSS.  Standby ...

jonathan-longe:
@infojunkie I'm not seeing the scrollbar locally.  Investigating ...

rjkuyvenhoven:
Popup scrolling depends on monitor resolution.  Also the bottom left admin menu impacts available area.  On one of my monitors I see the scrollbar when logged in as admin but works find at anonymous user.  My other monitor with higher resolution shows without scrollbar when logged in as admin.

jonathan-longe:
Thanks for the tip @rjkuyvenhoven  I wonder if we should make the video modal smaller?  I can easily hide the scrollbar, but if that make it impossible to press play in the video tool bar, we've ruined the user experience.  Thoughts?

jonathan-longe:
@infojunkie I found an example when the scrollbars is needed.  In iOS, if I click the thumbnail in portrait mode and then rotate the iPhone to landscape, I need the scrollbars to centre the video player in the modal window.  I think we'd need further testing before we can hide the scrollbars.

infojunkie:
Thanks for testing this. For now, let's leave eveerything as is.

PR #480, opened 2023-09-20
WBCAMS-224 update social icons

rjkuyvenhoven:


src/web/themes/custom/workbc/templates/modules/better-social-sharing-buttons.html.twig
```diff
@@ -15,6 +15,12 @@
       </a>
     {% endif %}
 
+    {% if items.services.x %}
```
infojunkie, line 18: I think you missed removing Twitter above.

jonathan-longe (approved):
Thanks for the opportunity to review.  Looks good

src/web/themes/custom/workbc/templates/modules/better-social-sharing-buttons.html.twig
rjkuyvenhoven: I left it there as a safety net in case items.services.twitter is provided instead of items.services.x

Was there a configuration or code change that changed the server from Twitter to X that is currently deployed to DEV & TEST and will be deployed to PROD tomorrow?

If so, I will remove the Twitter section in the twig file and the twitter icons.

infojunkie: Thanks for explaining. The latest version of `better_social_sharing_buttons` has converted Twitter => X, hence the discrepancy between the local version and what we see on DEV / TEST. I suggest to upgrade this module on this branch, to ensure your changes still hold in the latest version.

PR #483, opened 2023-09-22
WR-1162 Inconsistent pagination styling

rjkuyvenhoven:
There is no update for the Views Archive page as the pagination styling for this page is already correct (I tested this).


infojunkie (approved):
Lgtm, thanks!

PR #490, opened 2023-09-29
[WR-1733] theme half width icon cards

jonathan-longe:
Limit CSS changes for the half-width icon cards to `paragraph--type--action-cards-icon-1-2` class so changes do not affect the existing full-width cards. Screenshots below show how the cards appear after these changes:

Screenshot from 2023-09-29 11-49-40

Screenshot from 2023-09-29 11-49-53


infojunkie (approved):
Beautiful, thanks!

PR #493, opened 2023-10-04
WR-1002 image resizing text wrapping

rjkuyvenhoven:
enables CKEditor 5, CKEditor Media Resize & CKEditor 5 Media DL modules
configures Full HTML text format to use CKEditor 5 and Media Resize & DL modules
theme modified to prevent Media Resizing from being ignored


infojunkie:
Thanks @rjkuyvenhoven I'll test locally and feedback here.

infojunkie:
Working great for me!

Except that starting from a DEV snapshot, and attempting to sync the config, I get the error:
```
[error]  Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization.
Unable to uninstall the <em class="placeholder">Media entity download Link (CKEditor)</em> module because: Provides a filter plugin that is in use in the following filter formats: <em class="placeholder">Full HTML</em>. in Drupal\Core\Config\ConfigImporter->validate() (line 794 of /var/www/html/web/core/lib/Drupal/Core/Config/ConfigImporter.php). 
```
I had to restore `media_entity_dlcke: 0` in `core.extension.yml`, then sync the config, then manually uninstall `media_entity_dlcke`. We should make sure this doesn't happen.

rjkuyvenhoven:
ah ok. I'm not sure the best way forward.  First think is leave the module enabled. But when to disable it?  I think we'd have to leave the module enabled until these changes make it to PROD. Unless we want to have make manual uninstall part of the deploy process.  Unless you are aware of another way we could do this?

infojunkie:
I don't mind leaving it on, but we need to make sure it does not interfere with the new module. 

Another approach would be to remove the filter from the text format and try the sync procedure again.

infojunkie:
Working great now! I've tested:
- Image resizing
- Image alignment options
- Media link insertion

infojunkie (approved):
Ready for merge to `develop`!

PR #494, opened 2023-10-06
WBCAMS-252-129 event links

rjkuyvenhoven:
The Fullcalendar View module takes the last Views field and uses that to create a Description (des) field for each Event (I believe intended for a popup feature we are not using).  The data for each event (including node id) is available in the "Settings" variable passed by Drupal Behaviors.

I have reworked the Calendar view so that it now has two Global Text fields.  One (grid-view) containing the content displayed in the Calendar view and another (list-view) that contains the event description content displayed in below the Calendar when an event is clicked.  The list-view is the last field in the view configuration.

I've also added an div id to the grid-view  <div id="event-id-#> which is then used to target the correct event in the Settings data and an event is clicked.

As a result of splitting these two fields, event descriptions containing a link no longer results in a redirection.  Therefore, my fix is no longer needed and has been removed.

This PR also contains a fix for scrolling to the event description when an event is clicked (WBCAMS-129).

### DIFF
--- a/src/config/sync/views.view.content.yml
+++ b/src/config/sync/views.view.content.yml
@@ -97,7 +97,7 @@ display:
                 add_confirmation: false
                 label_override: 'Modify content'
                 get_bundles_from_results: true
-            11:
+            9:
               action_id: workbc_draft_node_action
               preconfiguration:
                 label_override: 'Create draft'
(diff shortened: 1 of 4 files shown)


infojunkie (approved):
I tested locally and the fix is working for me. Great job!

PR #504, opened 2023-10-18
WBCAMS 210 n/a note showing incorrectly

rjkuyvenhoven:
WBCAMS-210-Legislators-WorkBC


infojunkie:
Did you test with actual null values?

rjkuyvenhoven:
I tested it by programmatically setting one of the values to NULL and the N/A note displayed as expected.

I can look for an actual career profile that has a null value to confirm, if you'd prefer.

infojunkie (approved):
Thanks for testing both conditions. Lgtm!

PR #513, opened 2023-10-25
WR-1315 breadcrumb level 4 links not reaching the whole length of title

rjkuyvenhoven:
Modified the breadcrumb template so that the Save Profile div is only displayed if needed, thus no longer interfering with the links on the longer breadcrumbs.


infojunkie:
I'm not sure this is the right solution... what happens if the save profile is there AND the breadcrumbs are long? Can you replicate such a condition (e.g. by logging in to the Job Seeker account) ?

rjkuyvenhoven:
The clickable portion of Career & Industry profiles have specific length breadcrumbs that don't change.

Business-Building-and-Other-Support-Services-WorkBC
Air-pilots-flight-engineers-and-flying-instructors-WorkBC

rjkuyvenhoven:
Contractors-and-supervisors-machining-metal-forming-shaping-and-erecting-trades-and-related-occupations-WorkBC-2
Contractors-and-supervisors-machining-metal-forming-shaping-and-erecting-trades-and-related-occupations-WorkBC-1

infojunkie:
I'm afraid this will not fly in QA / UAT. We need some CSS work that ensures the "Save Profile" section is pushed down only when needed.

rjkuyvenhoven:
ok

rjkuyvenhoven:
CSS updated so "Save Profile" link appears on the same line if there's room and drops below if not.

Breadcrumb-1
Breadcrumb-2

PR #514, opened 2023-10-25
[WBCAMS-153] fix style of mobile industry overview table

jonathan-longe:
Fix industry overview table for mobile users.

Screenshot from 2023-10-25 14-08-14


src/web/themes/custom/workbc/templates/views/views-view-table--industry-profiles.html.twig
```diff
@@ -41,7 +41,6 @@
     'industry-overview-table',
   ]
 %}
-<div class="industry-overview-title">Industry Overview</div>
```
jonathan-longe, line 44: @rjkuyvenhoven I removed this title as it appears to duplicate the same text written above.  Okay with you?

infojunkie (approved):
Lgtm, thanks

src/web/themes/custom/workbc/templates/views/views-view-table--industry-profiles.html.twig
rjkuyvenhoven: With that removed I'm no longer seeing "Industry Overview" on either Mobile (what I'm seeing doesn't match the provided screenshot) or Desktop.  I'm not sure what the wireframes require for Mobile but it should be there for Desktop.

PR #517, opened 2023-10-26
WR-1738 update tables to right align data columns

rjkuyvenhoven:
Missed a couple tables that needed data columns right aligned.


infojunkie (approved):
Lgtm, thanks

PR #521, opened 2023-10-31
[WR-1512] Horizontal bar chart on BC Profile

infojunkie:


infojunkie:
screenshot-workbc docker localhost_8000-2023 10 31-00_13_18

infojunkie:
I made some progress with the styling of the B.C. Average, by creating a second chart with different styling. I managed to keep the alignment of the bars by hard-coding the chart width. 

I don't think it's feasible to add a grey background to the B.C. Average chart because it would include the legends, which would not conform to the wireframe and confuse the users.

I also note that the information tooltip next to the chart title is missing.

rjkuyvenhoven:
Thanks, that looks good to me.  

I'll add the tooltip.

infojunkie:
@rjkuyvenhoven can you please update the screenshot including the tooltip (preferably open) ? I want to show the client and get their approval early on.

rjkuyvenhoven:
wr-1512-chart-1

PR #524, opened 2023-11-01
[WR-1070] fix mobile tile display - version 2

jonathan-longe:
Further changes as requested by Sai

### DIFF
--- a/src/web/themes/custom/workbc/js/related-topics-carousel.js
+++ b/src/web/themes/custom/workbc/js/related-topics-carousel.js
@@ -21,7 +21,7 @@
           spaceBetween: 20
         },
         768: {
-          slidesPerView: 2,
+          slidesPerView: 1,
           spaceBetween: 20
         },
         992: {


infojunkie:
Can you include a screenshot from BrowserStack that this fixes the failure report?

jonathan-longe:
> Can you include a screenshot from BrowserStack that this fixes the failure report?

Good call. See below. Note, not all iPad's will display a single tile in portrait mode. The number of tiles displayed is dependent on screen width.

image

infojunkie (commented):
Looking at this change, I wonder how the 3 tiles were still showing in portrait mode. Shouldn't it have been 2 tiles only?

jonathan-longe:
> Looking at this change, I wonder how the 3 tiles were still showing in portrait mode. Shouldn't it have been 2 tiles only?

There were two tiles displayed previously, but they were broken up like this:  0.5 tile + 1 tile + 0.5 tile = 2 tiles.

infojunkie:
Thanks for the explanation - lgtm, thanks!

PR #525, opened 2023-11-01
WR-1058 selecting print duplicates sharing icons

rjkuyvenhoven:
print-social-sharing-icons
print-labour-market


infojunkie (approved):
Great fix!! Lgtm, thanks. 

Just to make sure: these screenshots are taken with the option "Include background images", correct?

rjkuyvenhoven:
Yes, these icons only show up with the option "include background images" checked. They don't show up at all when unchecked.

PR #526, opened 2023-11-01
WR-1570 update dashboard labels to display year on second line

rjkuyvenhoven:
wr-1570-bc-profile
wr-1570-region-profile
wr-1570-industry-profile
wr-1570-career-profile


infojunkie (changes requested):
It's bad localization practice to include HTML tags and variables (e.g. dates) in the `t()` function. I suggest you only enclose English text within `t()` and keep the rest outside. E.g. `return $this->t("Population - age 15+") . "<br>(" . date("M Y", $date1) . ")";`

rjkuyvenhoven:
Updated to remove tags and variables from t() function.  Label returned as markup array so tags are rendered as html.

infojunkie (approved):
Lgtm, thanks

PR #531, opened 2023-11-03
WR-1620 b.c. and regional profiles   lms tab   10 year unemployment rate graph display incorrect

rjkuyvenhoven:
B.C. Chart
wr-1620-bc-line-chart

Region Chart
wr-1620-region-line-chart


infojunkie (approved):
Lgtm, thanks

PR #532, opened 2023-11-03
WR 1629 mobile home page find post a job stacked

rjkuyvenhoven:
Reduced the nav-link padding for sm & xs media breakpoints

wr-1629-samsung-galaxy-S23-v13-0

wr-1629-iphone-13-v16-3


infojunkie (approved):
Thanks, this should be good enough.

PR #534, opened 2023-11-06
[WR-1082] show mobile menu for devices without hover

jonathan-longe:
This pull request changes the behaviour of our desktop and mobile menu.  In the past, the mobile menu would appear when the screen width was less than 1200 px.  Now, the mobile menu will appear if the screen width is less than 1200px OR the device doesn't recognize the "hover" attribute.  This change has been made to accommodate large tablets that couldn't use the existing desktop menu as it doesn't work with devices that can't hover over the menu items.


rjkuyvenhoven (approved):
lgtm

PR #540, opened 2023-11-15
Update to webform 6.2.0

rjkuyvenhoven:
Upgrade to Webform 6.2.0

NOTE:  The two instances where the filter format is changed to "Full HTML" are information fields for display on the forms, they are not fields the general public has access too.

I've included screen shots of the two fields in case they trigger a memory that there are tickets that supersede the wireframes.
 
webform620-Contact-Us-WorkBC
webform620-Request-a-Tour-Stop-WorkBC


infojunkie (approved):
Thanks for taking care of this! Lgtm

PR #541, opened 2023-11-16
WR-1613 update so field returns  "N/A" if null value is provided by SSoT

rjkuyvenhoven:
I also did a quick check to see if any other Extra Fields returned "not available" or similar and didn't see any others.


infojunkie (approved):
Lgtm, thanks

PR #544, opened 2023-11-16
WR-1706 interactive map

rjkuyvenhoven:
"Content" paragraph field added to BC, Region and Industry Profile content types. For these profiles the Interactive Map and Text components are available for use.


infojunkie (approved):
Lgtm, thanks

PR #551, opened 2023-11-23
 WR-1716-form-issues

rjkuyvenhoven:
enables Mobile Detect module, resolves 'illegal choice' & '0 selected' issues

### DIFF
--- a/src/web/modules/custom/workbc_custom/workbc_custom.module
+++ b/src/web/modules/custom/workbc_custom/workbc_custom.module
@@ -721,7 +721,7 @@ function workbc_custom_search_api_autocomplete_suggestions_alter(array &$suggest
     $params = $suggestion->getUrl()->getRouteParameters();
     $node = \Drupal\node\Entity\Node::load($params['node']);
     if ($node && isset($node->field_noc)) {
-      $markup = 'NOC ' . $node->field_noc->value . ' - ' . $suggestion->getLabel();
+      $markup = $suggestion->getLabel() . ' (NOC ' . $node->field_noc->value . ')';
       $markup = preg_replace("/(".$alter_params['user_input'].")/i", "<strong>$1</strong>", $markup);
       $render = array(
         '#type' => 'markup',
--- a/src/web/modules/custom/workbc_custom/workbc_custom.module
+++ b/src/web/modules/custom/workbc_custom/workbc_custom.module
@@ -190,7 +190,6 @@ function workbc_custom_form_views_exposed_form_alter(&$form, \Drupal\Core\Form\F
       $isMobile = $md->isMobile() && !$md->isTablet();
       $form['occupational_interest']['#attached']['drupalSettings'] = [
         'isMobile' => $isMobile,
-        'testMobile' => $md->isMobile(),
       ];
       if ($isMobile) {
         array_unshift($form['occupational_interest']['#options'], "All");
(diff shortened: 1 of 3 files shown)


src/web/modules/custom/workbc_custom/workbc_custom.module
```diff
@@ -185,7 +185,10 @@ function workbc_custom_form_views_exposed_form_alter(&$form, \Drupal\Core\Form\F
   }
 
   if (isset($form['occupational_interest'])) {
-    unset($form['occupational_interest']['#options']['All']);
+    $md = \Drupal::service('mobile_detect');
+    if ($md->isMobile()) {
```
infojunkie, line 189: Since we're using `isMobile()` on the server, I'd rather reuse the result on the client by sending a `drupalSettings` flag called `isMobile` instead of a possibly inconsistent method on the client. Alternatively, the client can know we're on mobile by searching for `All` in the options.

rjkuyvenhoven: Updated to pass `isMobile` status via `drupalSettings`

infojunkie:
Thanks I will test locally and feedback here.

rjkuyvenhoven:
Latest update moves selection handling to the backend.  Javascript is now only used to set All as the default selection, as I wasn't able to get this to work be setting the default value in the backend.

src/web/modules/custom/workbc_custom/workbc_custom.module
```diff
@@ -694,7 +721,7 @@ function workbc_custom_search_api_autocomplete_suggestions_alter(array &$suggest
     $params = $suggestion->getUrl()->getRouteParameters();
     $node = \Drupal\node\Entity\Node::load($params['node']);
     if ($node && isset($node->field_noc)) {
-      $markup = $suggestion->getLabel() . ' (NOC ' . $node->field_noc->value . ')';
+      $markup = 'NOC ' . $node->field_noc->value . ' - ' . $suggestion->getLabel();
```
infojunkie, line 724: I don't think this should be in here

rjkuyvenhoven: I thought I had merged master back into this branch.  Nevertheless this change has been removed.

src/web/themes/custom/workbc/js/mobile-chosen.js
```diff
@@ -2,52 +2,12 @@ let allSelected = true;
...
-          }
-
-        });
-
-        function isMobile() {
-          const regex = /Mobi|Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i;
-          return regex.test(navigator.userAgent);
+        if (drupalSettings.isMobile) {
+          var select = document.getElementById('edit-occupational-interest'); 
+          select.options[0].selected = true;
```
infojunkie, line 10: What happens here if "All" is not selected by the second option is selected?

rjkuyvenhoven: This happens once at page load.  I have not encountered a scenario where any other selection has already been made.
However, if a second option were selected both would be highlighted and that option would be ignored when Apply is clicked.

In the interest of covering all bases I can update this so all other options have selected set to false if you'd like.

infojunkie: If I understand correctly, this only happens at full-page load, not in the Ajax refresh? If that's the case, it makes sense as-is.

rjkuyvenhoven: yes, only on full-page load

infojunkie (approved):
Thanks for a very thorough fix! Lgtm

PR #554, opened 2023-11-28
[WR-1423] fix labour market statistics table styling rev2

jonathan-longe:
Use equal column width for LMS "Employment by Region" table to resolve misunderstanding with the business.

Screenshot from 2023-11-27 16-00-10


infojunkie (approved):
Lgtm, thanks

PR #555, opened 2023-11-28
WR-1151 add <figure> tag to allowed tag list.

rjkuyvenhoven:
The content was being stripped out due to CKEditor 5 allowed tags.  Although we allow Drupal to render all tags CKEditor still works off a list of specific allowable tags.  I've added the <figure> tag to the list.

Once the latest code is deployed to TEST the client will revert back to the previous version of the page and then apply the "Content Source" style the appropriate text and the image will not be lost.


infojunkie (approved):
Lgtm, thanks. 

Until next deployment, is it possible to edit the tags via Admin UI and perform the fix ourselves on TEST ?

rjkuyvenhoven:
I think so.

PR #567, opened 2023-12-11
[WR-1002] Media resize without limit tags

infojunkie:
Allow the CKE5 filter plugin **Resize media images** to function without the filter plugin **Allow all HTML tags & attributes and correct faulty HTML**.

We're keeping the latter around because deleting it would need to be done in a second deployment step, to avoid crashing the PHP app server.


src/web/modules/custom/ckeditor_media_resize/js/ckeditor5_plugins/mediaResize/src/mediaresizeediting.js
```diff
@@ -5,72 +5,95 @@ const MEDIA_WIDTH_DATA_ATTRIBUTE = 'data-media-width';
...
-					viewWriter.removeStyle( 'width', mediaView );
+          viewWriter.addClass( 'image_resized', mediaView );
+        } else {
+          viewWriter.removeStyle( 'width', mediaView );
+          viewWriter.removeAttribute( MEDIA_WIDTH_DATA_ATTRIBUTE, mediaView );
+          viewWriter.removeClass( 'image_resized', mediaView );
+        }
+      } )
+
+      dispatcher.on( `attribute:drupalMediaIsImage:${ elementType }`, ( evt, data, conversionApi ) => {
```
infojunkie, line 74: This is the actual fix: We react to this CKE5 event by setting the width attribute from the `data.item` element into the `drupal-media` view element.

PR #570, opened 2023-12-14
WBCAMS-168 summary field for all page templates remove hide this field

rjkuyvenhoven:
Home-WorkBC
Edit-Blog-Career-Services-for-Post-Secondary-Students-and-Alumni-WorkBC
Edit-Page-Indigenous-Skills-and-Employment-Training-WorkBC


src/web/themes/custom/workbc/workbc.theme
```diff
@@ -222,6 +222,29 @@ function workbc_preprocess_node(&$variables) {
...
+  if ($variables['node']->bundle() === "blog") {
+    $featured = "";
+    $node = $variables['node'];
+    if ($node->hasField('body')) {
+      if (!empty($node->get('body')->summary)){
+        $featured = $node->get('body')->summary;
+      }
+      else {
+        if ($node->hasField('field_related_topics_blurb') && !empty($node->get('field_related_topics_blurb')->value)) {
+          $featured = strip_tags($node->get('field_related_topics_blurb')->value);
```
infojunkie, line 234: Why are we converting the blurb to plain text?

src/web/themes/custom/workbc/workbc.theme
```diff
@@ -222,6 +222,29 @@ function workbc_preprocess_node(&$variables) {
...
+        }
+        else {
+          if (!empty($node->get('body')->value)) {
+            $text = strip_tags($node->get('body')->value);
```
infojunkie, line 238: Previously, the node blog teaser used the full `content.body` value. Since this is a fallback, do we need to truncate and strip tags here?

src/web/themes/custom/workbc/templates/content/node--blog--teaser.html.twig
```diff
@@ -71,7 +71,7 @@
 {% set image = content.field_image_media %}
 {% set date = content.extra_field_workbc_published_on %}
 {% set title = content.extra_field_workbc_page_title %}
-{% set body = content.body|render|striptags|trim %}
+{% set body = featured_text|render|striptags|trim %}
```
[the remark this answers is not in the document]
rjkuyvenhoven: Yes, this template is used by the Blog Featured block view which displays a single (most recent) Blog post on the front page.

src/web/themes/custom/workbc/workbc.theme
rjkuyvenhoven: It used the Blog Manage Display Teaser view which configured the Body field to use the "Summary or trimmed" format.  So we need to truncate but we don't need to  strip tags.

infojunkie: We may need to strip the tags before truncating, unless truncating is smart enough to take tags into account.

rjkuyvenhoven: Good point. I'm not sure truncate is smart enough.  Will test and update code as needed.

infojunkie:
Looking good so far, thanks. I am waiting to hear from the product team on where the summary for non-blog content is expected to appear. It is unnecessary to expose a field that is used nowhere.

rjkuyvenhoven:
It should be noted that the Summary field may be used as Related Topics text if Related Topics Blurb or Hero Text isn't available.

infojunkie:
Thanks, so can you confirm that the summary field appears for all content types that can be used for related topics plus blogs?

rjkuyvenhoven:
This PR makes visible the Summary field for content types with a Body field, these include Blog, Events, News, Page and Success Story. All these content types can be used as a Related Topic.  Other content types available for use as a Related Topic do not have a Body field.

infojunkie (approved):
Thanks for the explanation. Lgtm!

PR #577, opened 2023-12-18
WR-1757-1765-1784 regional profile l3 page   label   text contrast is insufficient

rjkuyvenhoven:
Accommodation-and-Food-Services-WorkBC


infojunkie (approved):
Lgtm, thanks

PR #581, opened 2023-12-19
 WR-1783 image-has-neither-an-alt-attribute-nor-an-ARIA-label-or-title

rjkuyvenhoven:


infojunkie (approved):
As long as it passes the a11y checker, it looks good to me.

PR #596, opened 2024-01-26
WR-1801 interactive map

rjkuyvenhoven:


infojunkie (approved):
Great solution, thanks!

PR #611, opened 2024-02-27
WBCAMS-349 interactive maps

rjkuyvenhoven:
removed alt text to prevent screen reader from describing the lines.  The alt text is left as an empty string to avoid triggering an accessibility assessment error.


infojunkie (approved):
Lgtm, thanks

PR #612, opened 2024-02-27
NOC-232 update search indexes

rjkuyvenhoven:


infojunkie:
Thanks. What are the steps needed for this configuration change to take effect and the indexes to include this new field?

rjkuyvenhoven:
After the configuration takes effect all items for each search index will be queued for reindexing and will be reindexed during cron runs.

infojunkie (approved):
Lgtm, thanks. Please add deployment instructions to the ticket.

PR #616, opened 2024-02-29
NOC-298 Sort industry data by name rather than key to ensure correct display order

rjkuyvenhoven:


infojunkie (commented):
Is there a specific key that is causing issues, e.g. `other_private_services` ?

rjkuyvenhoven:
Yes, that is the one

infojunkie:
Thanks. As we agreed offline, it's safer to keep the key as is and sort the names as per your approach.

infojunkie (approved):

PR #619, opened 2024-03-05
Update JB commit hash 0e5186c

jonathan-longe:
Simple change - updating the Job Board commit hash


infojunkie (changes requested):
This needs to be again branches ams or noc depending on which tickets it addresses

PR #625, opened 2024-03-06
WBCAMS-343 bc profile  lms  employment by region table styling

rjkuyvenhoven:
High-Opportunity-Occupations-WorkBC


infojunkie (approved):
Short and sweet thanks!

PR #638, opened 2024-04-08
WBCAMS-14 link to Career Trek site on associated right-side video

rjkuyvenhoven:
To validate post update:
http://workbc.docker.localhost:8000/admin/reports/career-trek-validation

CareerProfile-21233-WorkBC


infojunkie:
Thanks, working well for me. Some comments:
- I added a default value for the CT URL on local
- We need to wait for the CAREERTREK_URL variable to be added on DEV to merge
- There is an updated CT dataset dropping soon which will contain the "official" URL for each episode - we'll update the migration here to use that instead

infojunkie (approved):
Looks great! Everything checks out, migration report clean.

PR #648, opened 2024-05-10
[NOC-356]  Read industries metadata from SSOT

infojunkie:


infojunkie:
@rjkuyvenhoven I did some further refactoring - please test on your local machine when you get the chance.

One last thing I'd like to discuss with you is handling LMM industries which as you know are not 1:1 mapping to industry profiles. Right now, the mapping is done in 2 places:
- `ssotIndustryLMMKey()` when displaying LMM data on the Industry Profile page.
- `LabourMarketEmploymentIndustriesTable::getIndustryHighlights()` when displaying the LMM page itself - the only needed data here are the labels.

The first one can be handled more elegantly by simply renaming the LMM industry columns in the SSOT `monthly_labour_market_updates` endpoint to match the `industries` keys.

The second one is more problematic, because an LMM industry such as `wholesale_and_retail_trade` does not have an industry profile equivalent. I am reluctant to add an entry for it in the `industries` table, but it would still be great to have a way to read these labels from SSOT.  I am thinking to read these labels from the SSOT schema of the `monthly_labour_market_updates` endpoint.

What do you think?

infojunkie:
I've updated SSOT with the changes outlined above - to test, please refresh your SSOT with the dump in this branch.

rjkuyvenhoven (approved):
Tested on my local dev environment. 

lgtm

PR #657, opened 2024-07-08
WBCAMS-500 Add metatag description field to Blog, News, Page and Success Story content types…

rjkuyvenhoven:
Add metatag description field to Blog, News, Page and Success Story content types, configure Metatag module to use metatag description field to provide content for description metatag.

The new Metatag Description field has been placed below the Tags field on the edit screen for each of the content types.

Edit-Blog-Career-Services-for-Post-Secondary-Students-and-Alumni-WorkBC (1)

Blog-Career-Services-for-Post-Secondary-Students-and-Alumni-WorkBC-metatag-description


src/config/sync/core.entity_form_display.node.success_story.default.yml
```diff
@@ -123,6 +124,14 @@ content:
       size: 60
       placeholder: ''
     third_party_settings: {  }
+  field_test:
+    type: string_textfield
+    weight: 26
+    region: content
+    settings:
+      size: 60
```
infojunkie, line 132: is this 60 or 160?

src/config/sync/field.storage.node.field_test.yml
```diff
@@ -0,0 +1,21 @@
...
+status: true
+dependencies:
+  module:
+    - node
+id: node.field_test
+field_name: field_test
+entity_type: node
+type: string
+settings:
+  max_length: 255
```
infojunkie, line 12: should this be 160?

src/web/modules/custom/workbc_custom/workbc_custom.module
```diff
@@ -484,6 +484,48 @@ function workbc_custom_search_api_autocomplete_suggestions_alter(array &$suggest
...
+    if ($tag[1] == 'description') {
+      $found = true;
+    }
+  }
+
+  if (!$found) {
+    $node = \Drupal::routeMatch()->getParameter('node');
+    $description = $node->title->value;
+    if ($node->hasField('body')) {
+      if (!empty($node->get('body')->value)) {
```
infojunkie, line 506: should we examine the summary too?

src/config/sync/core.entity_form_display.node.success_story.default.yml
rjkuyvenhoven: I'll need to dig into this. It should be 160 and my local is configured for 160.

src/config/sync/field.storage.node.field_test.yml
rjkuyvenhoven: This is a test field that should have been deleted. Will update the PR.

src/config/sync/core.entity_form_display.node.success_story.default.yml
infojunkie: Maybe it refers to the HTML attribute `size` as opposed to the max length https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/size

src/web/modules/custom/workbc_custom/workbc_custom.module
rjkuyvenhoven: That seems like a reasonable option.  However, the acceptance criteria says:

"If the "Metatag Description" field is empty, ensure the module defaults to using the first 160 characters of the page content as the meta description."

src/config/sync/core.entity_form_display.node.success_story.default.yml
rjkuyvenhoven: Yes, that could be. I thought I had changed them all but may have missed one.  I'm looking into this now.

src/web/modules/custom/workbc_custom/workbc_custom.module
infojunkie: Before we stick to the letter of the ticket, are there content types where the metatag description is active, where the body summary is also active? If yes, it would be better to discuss the spec to avoid unmet expectations.

src/config/sync/core.entity_form_display.node.success_story.default.yml
rjkuyvenhoven: This config was for the test field that has been deleted.

src/web/modules/custom/workbc_custom/workbc_custom.module
rjkuyvenhoven: Yes, there is.  I will ask in the ticket.

rjkuyvenhoven: I've updated to code to use the body summary if one is available

src/web/modules/custom/workbc_custom/workbc_custom.module
```diff
@@ -484,6 +484,53 @@ function workbc_custom_search_api_autocomplete_suggestions_alter(array &$suggest
...
+    $node = \Drupal::routeMatch()->getParameter('node');
+    $description = $node->title->value;
+    if ($node->hasField('body')) {
+      if (!empty($node->get('body')->summary)){
+        $description = \Drupal\Component\Utility\Unicode::truncate($node->get('body')->summary, 160, TRUE, FALSE);
+      }
+      else {
+        if (!empty($node->get('body')->value)) {
+          $text = strip_tags($node->get('body')->value);
+          $text = str_replace(array("\r", "\n"), ' ', $text);
```
infojunkie, line 512: I think the `str_replace` is not needed because the `preg_replace` takes care of newlines.

rjkuyvenhoven: yes, this was unneeded and has been removed.

infojunkie (approved):

PR #665, opened 2024-08-26
[WBCAMS-521,541,542,543] Career Trek / Video Library changes

infojunkie:


src/config/sync/block.block.workbcexplorecareersblock.yml
```diff
@@ -23,7 +23,7 @@ settings:
   title: 'Career Profiles'
   body: 'Explore our 500 career profiles to find job expectations, education requirements, salary and more.'
   title_1: 'Search by keyword(s)'
-  body_1: 'Look up a career profile by entering a keyword(s) from the job title, or by entering the five-digit NOC number.'
+  body_1: 'Look up a career profile by entering a keyword(s) from the job title, or by entering the four-digit NOC number.'
```
infojunkie, line 26: @rjkuyvenhoven I think there are unwanted config files that snuck in here. Please delete the config files that are not related to this CT work.

src/web/modules/custom/workbc_custom/workbc_custom.post_update.php
```diff
@@ -1032,4 +1034,95 @@ function workbc_custom_post_update_14_wbcams_migration(&$sandbox = NULL) {
...
+
+  $media = Media::load($video->mid);
+  if ($media) {    
+    if (is_null($media->field_weight->value)) {
+      $media->field_weight = 0;
+      $media->save();
+      $message = "Remote Video: " . $media->name->value . " - weight set to 0";
+    }
+    else {
+      $message = "Remote Video: " . $media->name->value . " - weight already set";
```
infojunkie, line 1118: I don't understand why the weight would be already set, since we just introduced this field. Instead of this whole migration, can't we just rely on the default value of this field which is set to zero?

rjkuyvenhoven: The default value for this field for existing records is NULL.  Views displays NULL values first so the field needs a default value to be useful.

Yes, weight already set isn't strictly necessary. However, in my testing I had set the weight value for some videos before creating the post_update to set the default weight, so I added it to confirm the post update was working as expect.

infojunkie: Sounds good, let's keep it if it is stable on DEV

rjkuyvenhoven (approved):
lgtm

PR #688, opened 2024-10-22
WBCAMS-688 drupal 10 update ckeditor media resize

rjkuyvenhoven:
modules/custom/ckeditor_media_resize

CKEditor Media Resize module version1.0.0-beta2 with patch applied 

I've added the patch to the patches folder and added the appropriate info to composer.json.  However, these are currently ignored.


infojunkie (approved):
Looks good but please make sure to rebase to the latest ams because the D10 upgrade branch is gone.

PR #692, opened 2024-10-28
WBCAMS-688 module upgrades

rjkuyvenhoven:
- freezes core to 10.3.6
- uninstalls actions_permissions module


infojunkie (approved):
That's fine, thanks. We still upgrade modules manually at each deployment so we'll have the opportunity to upgrade Drupal to 10.4+ then.

PR #695, opened 2024-11-13
WBCAMS-600 hoo wage filter

rjkuyvenhoven:
I looked into doing some custom filtering but it didn't look feasible and simply setting hourly wage for $50+ option to 2500 accomplishes the task based on the assumption no hourly rate will be as high as $2,500/hour or any salary as low as $2,500/year


infojunkie (approved):
If you know of an occupation that pays >= $2500 / hr please let me know! Otherwise lgtm :-D

rjkuyvenhoven:
> If you know of an occupation that pays >= $2500 / hr please let me know! 

I'm not aware of one but I'll keep my eyes open...

PR #702, opened 2024-11-30
[WBCAMS-719] LMO Report 2024 content type

infojunkie:


src/web/themes/custom/workbc/templates/field/field--paragraph--extra-field-lmo-report-2024-job-openings-regions-table.html.twig
infojunkie, line 1: Since we're injecting the table markup in the code directly, there's no need to inject the map markup in a separate theme template. Let's keep them in the same place.

rjkuyvenhoven: The include file '@workbc/includes/interactive-map.html.twig' is the same file we use for all the interactive maps on the site.

PR #704, opened 2024-12-02
WorkBC24-R7 TESTv3

infojunkie:


infojunkie:
@rjkuyvenhoven looks like GitHub Copilot is complaining about a few things in Bootstrap... can you please take a look and advise on a minimal fix? A patch would be fine.

rjkuyvenhoven (approved):

PR #716, opened 2025-01-16
WBCAMS-823 re sized images not adjusting for mobile

rjkuyvenhoven:
iPhone-16-v18-2-01-16-2025_02_06_PM


infojunkie (approved):
Fix looks simple and safe. Thanks!

PR #719, opened 2025-01-23
[WBCAMS-809] video dont show flag

rjkuyvenhoven:


src/config/sync/field.field.media.remote_video.field_exclude_from_video_page.yml
```diff
@@ -0,0 +1,21 @@
...
+id: media.remote_video.field_exclude_from_video_page
+field_name: field_exclude_from_video_page
+entity_type: media
+bundle: remote_video
+label: "Don't show on Career Trek page"
+description: ''
+required: false
+translatable: false
+default_value: {  }
+default_value_callback: ''
```
infojunkie, line 17: can we use `default_value`or the `default_value_callback`  for existing videos instead of a new data migration? migrations make me nervous!

infojunkie (commented):
Left a question

src/config/sync/field.field.media.remote_video.field_exclude_from_video_page.yml
rjkuyvenhoven: I've added a default value for the sake of completeness.   However, these options only work when adding or editing content.  It doesn't impact existing content.  Views only allows us to filter on True/False and without post update function the value of the flag for existing content would be NULL.

infojunkie: Sounds good. Please add to the ticket a ballpark estimate of how long it takes for this data update to run (e.g. < 1  minute, 1-5 minutes, 5-15 minutes, etc.)

infojunkie (approved):
Looks good, thanks

PR #720, opened 2025-01-23
[WBCAMS-809] use deploy hook for migrating data

rjkuyvenhoven:
modified to use deploy hook so data updates occur post configuration update.


infojunkie (approved):
Tested and working locally. Thanks!

PR #764, opened 2025-05-18
WorkBC25-R1 TESTv2

infojunkie:


src/web/themes/custom/workbc/workbc.theme
```diff
@@ -410,25 +451,6 @@ function workbc_preprocess_field(&$variables) {
   }
 }
 
-/*
- * Implements hook_theme_suggestions_HOOK_alter()
- */
```
infojunkie, line 415: These were wrongly removed - need to restore them.

PR #873, opened 2026-06-16
[WBCAMS-1997] alt value for images and icons

rjkuyvenhoven:


src/web/modules/custom/workbc_custom/workbc_custom.deploy.php
```diff
@@ -108,3 +108,59 @@ function workbc_custom_deploy_1717_import_centre_regions(&$sandbox = NULL) {
...
+      }
+      $media->save();
+      $message = $image->bundle . ": " . $image->mid . " - alt text changed to empty string.";
+    }
+    else {
+      $message = $image->bundle . ": " . $image->mid . " - alt text unchanged.";
+    }
+  }
+  else {
+    $message = $search . " - Media " . $record->mid . " not found";
```
infojunkie: I think this needs to be `$image` not `$record`

infojunkie (changes requested):
One change needed

infojunkie (approved):

PR #879, opened 2026-06-18
[WBCAMS-1998] resolve quick reads duplicate card issue

rjkuyvenhoven:
This issue is the result of multiple Success Stories with the same Published date and PostgreSQL LIMIT and OFFSET not guaranteeing any default order without unique sort by values.  The problem is solved by adding Node ID to View Sort criteria.

I added a note to the field label to make clear using the ID as part of the Sort does serve a purpose.


infojunkie (approved):
Good debugging!

infojunkie (approved):
Good debugging!

PR #881, opened 2026-07-09
Give Job Board's hidden checkbox inputs a 16px hit box for voice control

bidyashish:
## Problem (WEP a11y: voice control cannot interact with checkboxes)
On `workbc.ca/account#/login` and `#/register`, Windows Voice Access reports *"We can't find check box on youth"*. Root cause is in this theme's `_job_board.scss`:

```scss
app-root input[type=checkbox], input[type=radio] {
  appearance: none; opacity: 0; width: 0; position: absolute;
}
```

This collapses every native input inside the embedded Job Board SPAs to a **0×0** box. Screen readers can still announce it (it keeps an accessible name), but voice control needs a real bounding rectangle to click — and there is none. Aria fixes on the Job Board side could not solve this.

## Fix
- Keep the native input invisible (`opacity:0`) but give it a `16px × 16px` box (+`margin:0`) so it sits exactly over the SVG checkbox artwork — the WebAIM-recommended custom-checkbox pattern.
- `pointer-events: none` on the decorative `.checkbox-icons` overlay: the SPA raises it above the input on `:focus`, which otherwise swallows the mouseup mid-click and cancels the toggle.
- No `!important` on the input sizing so the existing `.job-alerts-inner input[type=radio]` restore-to-native override keeps winning.
- `css/style.css` patched in lockstep (compiled artifact; only the one rule touched).

## Verification
Headless-Chromium against the live dev pages with this patched `style.css` substituted via request interception: checkbox bounds go 0×0 → 16×16 and real mouse clicks toggle reliably (`false → true → false → true`) on register (`#isYouth`, `#JobSeekerIsApprenticeship`) and login (`#staySignedIn`).

Companion defense-in-depth change on the Job Board side: bcgov/workbc-jb#360.


infojunkie:
Please target the `ams` branch to allow this to go to DEV / QA first.

PR #886, opened 2026-07-16
[WBCAMS-2086] decrease menu dropdown white space.

rjkuyvenhoven:
The field length of the navigation blurb field can not be changed in the UI so it is done programmatically.

The `text-overflow: ellipsis;` line was removed because it isn't needed.  `-webkit-line-clamp` handles adding the ellipsis if needed.

`max-height: 375px` is used instead of a line clamp for the Quick Links area because `-webkit-line-clamp` requires `display: -webkit-box` and this div is using `display: flexbox;` for formatting purposes.


src/web/modules/custom/workbc_custom/workbc_custom.post_update.php
```diff
@@ -1503,3 +1504,65 @@ function workbc_custom_post_update_1997_media_image_alt_text(&$sandbox = NULL) {
...
+ * Updates the length of a text field which already contains data.
+ *
+ * @param string $entity_type_id
+ *   The entity type ID.
+ * @param string $field_name
+ *   The field name to update.
+ * @param int $new_length
+ *   The new maximum length for the field.
+ */
+function _change_text_field_max_length(string $entity_type_id, string $field_name, int $new_length): void {
```
infojunkie: Sounds like too much code...

How about changing the form element's `maxlength` attribute? https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/maxlength

infojunkie (changes requested):
I hope we can reduce the amount of new code.

infojunkie (approved):

PR #889, opened 2026-07-21
[WBAMS-2056] front page hero image

rjkuyvenhoven:
no cache on front page hero image to allow for randomization


infojunkie (approved):
Good catch

